PR #3, opened 2019-05-27
Initial implementation of UnpublishedMutableData

lionel1704:
This PR contains the inital implementation of unpublished mutable data as specified by [RFC 54](https://github.com/maidsafe/rfcs/blob/master/text/0054-published-and-unpublished-mutable-data/0054-published-and-unpublished-mutable-data.md)


src/mutable_data.rs
```diff
@@ -0,0 +1,135 @@
...
+    Read,
+    Insert,
+    Update,
+    Delete,
+    ManagePermissions,
+}
+
+/// Mutable data that is unpublished on the network. This data can only be fetch be the owners / those in the permissions fiedls with `Permission::Read` access.
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+pub struct UnpublishedMutableData {
```
nbaksalyar: I'd say let's just leave it at just `MutableData` since it is not a application developer facing API. It is implied that MutableData is unpublished, so this will be an unnecessary qualification here.

src/response.rs
```diff
@@ -0,0 +1,24 @@
...
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use crate::errors::ClientError;
+use crate::mutable_data::UnpublishedMutableData;
+use crate::MessageId;
+
+/// Rpc responses from vaults.
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+pub enum Response {
```
nbaksalyar: Can we make it `pub enum Response<ErrorType>` and replace `ClientError` with `ErrorType`?
Then we don't even need to define the error type in this crate.

src/mutable_data.rs
lionel-faber: I left the prefix in there to differentiate this from the existing data type in routing. But I guess we can always rename the import to show the difference :+1:

mrcnski: I also proposed this change to the RFC in case you guys haven't seen it:
https://github.com/maidsafe/rfcs/pull/309

PR #4, opened 2019-05-27
Initial implementation of UnpublishedImmutableData

m-cat:
Closes https://github.com/maidsafe/safe_client_libs/issues/808


src/immutable_data.rs
```diff
@@ -0,0 +1,67 @@
...
+use tiny_keccak;
+
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+pub struct UnpubImmutableData {
+    /// Contained ImmutableData.
+    data: Vec<u8>,
+    /// Contains a set of owners of this data. DataManagers enforce that a
+    /// DELETE or OWNED-GET type of request is coming from the
+    /// MaidManager Authority of the owners.
+    owners: PublicKey,
```
nbaksalyar, line 20: This should be [`threshold_crypto::PublicKeySet`](https://docs.rs/threshold_crypto/0.3.1/threshold_crypto/struct.PublicKeySet.html) (i.e. `BLS-PublicKey`).
If `threshold_crypto` is not included as a dependency, could you pls do this as well?

mrcnski: Sure, I just saw it was commented out in Cargo.toml. I'll add it to the crate.

PR #5, opened 2019-05-28
Move ImmutableData from routing to safe-nd

m-cat:
Part 1 of https://github.com/maidsafe/safe_client_libs/issues/809

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,12 +10,14 @@ authors = ["MaidSafe Developers <dev@maidsafe.net>"]
 edition = "2018"
 
 [dependencies]
-hex = "~0.3.1"
-maidsafe_utilities = "~0.17.0"
-rand = "~0.4.2"
+maidsafe_utilities = "~0.18.0"
 rust_sodium = "~0.10.2"
 serde = "~1.0.37"
 serde_derive = "~1.0.27"
 threshold_crypto = "~0.3.1"
 tiny-keccak = "~1.4.2"
+
+[dev-dependencies]
+hex = "~0.3.1"
+rand = "~0.4.2"
 unwrap = "~1.2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,8 +10,12 @@ authors = ["MaidSafe Developers <dev@maidsafe.net>"]
 edition = "2018"
 
 [dependencies]
-threshold_crypto = "~0.3.1"
+hex = "~0.3.1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -10,8 +10,12 @@ authors = ["MaidSafe Developers <dev@maidsafe.net>"]
 edition = "2018"
 
 [dependencies]
-threshold_crypto = "~0.3.1"
+hex = "~0.3.1"
+maidsafe_utilities = "~0.17.0"
+rand = "~0.4.2"
```
nbaksalyar: Should be in `[dev-dependencies]`. And probably we can use `~0.6.4` here?

Cargo.toml
```diff
@@ -10,8 +10,12 @@ authors = ["MaidSafe Developers <dev@maidsafe.net>"]
 edition = "2018"
 
 [dependencies]
-threshold_crypto = "~0.3.1"
+hex = "~0.3.1"
+maidsafe_utilities = "~0.17.0"
```
nbaksalyar: We published 0.18.0 recently, so could you use it here too please?
```suggestion
maidsafe_utilities = "~0.18.0"
```

Cargo.toml
```diff
@@ -10,8 +10,12 @@ authors = ["MaidSafe Developers <dev@maidsafe.net>"]
 edition = "2018"
 
 [dependencies]
-threshold_crypto = "~0.3.1"
+hex = "~0.3.1"
```
nbaksalyar: Should be in `[dev-dependencies]`.

Cargo.toml
mrcnski: Looks like `rand = "~0.6.4"` doesn't work because maidsafe_utilities 0.18.0 uses 0.3.15:

```
error[E0599]: no method named `gen_iter` found for type `maidsafe_utilities::seeded_rng::SeededRng` in the current scope
   --> src/immutable_data.rs:159:25
    |
159 |         let value = rng.gen_iter().take(len).collect();
    |                         ^^^^^^^^
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
            `use rand::Rng;`

warning: unused import: `rand::Rng`
   --> src/immutable_data.rs:111:9
    |
111 |     use rand::Rng;
    |         ^^^^^^^^^
    |
    = note: #[warn(unused_imports)] on by default
```

PR #8, opened 2019-05-31
Proper implementation of the mutable data struct

lionel1704:
- Implement published and unpublished mutable data structs as per the RFC
- Introduce `DataError` and `EntryError` enums
- Implement `Debug` trait for `Request` and `Response` enums

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,4 +1,4 @@
-// Copyright 2018 MaidSafe.net limited.
+// Copyright 2019 MaidSafe.net limited.
 //
 // This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
 // Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,0 +1,104 @@
+// Copyright 2018 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use std::collections::BTreeMap;
+use std::error::Error;
+use std::fmt::{self, Display, Formatter};
+
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+pub enum DataError {
+    /// Access is denied for a given requester
+    AccessDenied,
+    /// /// SAFE Account does not exist for client
+    NoSuchAccount,
+    /// Attempt to take an account network name that already exists
+    AccountExists,
+    /// Requested data not found
+    NoSuchData,
+    /// Attempt to create a mutable data when data with such a name already exists
+    DataExists,
+    /// Requested entry not found
+    NoSuchEntry,
+    /// Exceeded a limit on a number of entries
+    TooManyEntries,
+    /// Some entry actions are not valid.
+    InvalidEntryActions(BTreeMap<Vec<u8>, EntryError>),
+    /// Key does not exist
+    NoSuchKey,
+    /// The list of owner keys is invalid
+    InvalidOwners,
+    /// Invalid version for performing a given mutating operation. Contains the
+    /// current data version.
+    InvalidSuccessor(u64),
+    /// Invalid Operation such as a POST on ImmutableData
+    InvalidOperation,
+    /// Network error occurring at Vault level which has no bearing on clients, e.g. serialisation
+    /// failure or database failure
+    NetworkOther(String),
+}
+
+impl<T: Into<String>> From<T> for DataError {
+    fn from(err: T) -> Self {
+        DataError::NetworkOther(err.into())
+    }
+}
+
+impl Display for DataError {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match *self {
+            DataError::AccessDenied => write!(f, "Access denied"),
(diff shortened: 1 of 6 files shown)


src/errors.rs
```diff
@@ -0,0 +1,104 @@
+// Copyright 2018 MaidSafe.net limited.
```
mrcnski: Minor point but this should be 2019 (not sure how I even noticed this).

src/mutable_data.rs
```diff
@@ -7,58 +7,117 @@
...
 // of the SAFE Network Software.
+
+use crate::errors::{DataError, EntryError};
 use crate::XorName;
+use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, BTreeSet};
+use std::mem;
 use std::vec::Vec;
-use threshold_crypto::PublicKey;
+use threshold_crypto::*;
```
lionel-faber, line 17: Should avoid using `*`
Will change this.

src/mutable_data.rs
```diff
@@ -7,58 +7,117 @@
...
 
-/// Used to differentiate between conflict-free and conflict aware variants.
+/// Mutable data that is unpublished on the network. This data can only be fetch be the owners / those in the permissions fields with `Permission::Read` access.
```
nbaksalyar: Typo:
```suggestion
/// Mutable data that is unpublished on the network. This data can only be fetch by the owners / those in the permissions fields with `Permission::Read` access.
```

src/mutable_data.rs
```diff
@@ -7,58 +7,117 @@
...
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize, Debug)]
+pub struct Value {
+    /// Actual data.
+    data: Vec<u8>,
+    /// SHALL be incremented sequentially for any change to `data`.
+    version: u64,
 }
 
-/// Mutable data that is unpublished on the network. This data can only be fetch be the owners / those in the permissions fiedls with `Permission::Read` access.
+/// Mutable data that is unpublished on the network. This data can only be fetch be the owners / those in the permissions fields with `Permission::Read` access.
```
nbaksalyar: Typo here too :)
```suggestion
/// Mutable data that is unpublished on the network. This data can only be fetch by the owners / those in the permissions fields with `Permission::Read` access.
```

src/mutable_data.rs
```diff
@@ -7,58 +7,117 @@
...
     /// coming from the MaidManager Authority of the Owner.
-    /// Currently limited to one owner to disallow multisig.
     owners: PublicKey,
 }
 
-impl MutableData {
+/// Set of user permissions.
+#[derive(Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord, Serialize, Deserialize, Default)]
+pub struct PermissionSet {
+    permissions: BTreeSet<Action>,
```
nbaksalyar, line 67: Eventually it'd be better to replace `PermissionSet` with just `BTreeSet<Action>`, but now for compatibility it's a good idea to leave it as it is. Nice one :+1:

src/mutable_data.rs
```diff
@@ -68,48 +127,537 @@ impl MutableData {
...
+            return false;
+        }
+
+        self.owners = new_owner;
+        self.version = version;
+        true
+    }
+
+    pub fn is_action_allowed(&self, requester: PublicKey, action: Action) -> bool {
+        if self.owners == requester {
```
nbaksalyar, line 372: This wouldn't be correct really with BLS, since `PublicKey` can be comprised of multiple keys. So it should take a signature for an owner instead. But we'll deal with it as a separate task, so OK to leave it as such for now.

src/mutable_data.rs
```diff
@@ -68,48 +127,537 @@ impl MutableData {
...
+            return true;
+        }
+        match self.permissions.get(&requester) {
+            Some(perms) => perms.clone().is_allowed(action),
```
nbaksalyar: I don't think we need to do a `clone` here. `is_allowed` borrows `&self` as a reference, so it shouldn't be really required.

src/mutable_data.rs
```diff
@@ -68,48 +127,537 @@ impl MutableData {
...
 }
 
-/// Subject of permissions
-#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
-pub enum User {
-    /// Permissions apply to a single public key.
-    Key(PublicKey),
-    /// Permissions apply to anyone.
-    Anyone,
+/// Implements functions which are COMMON for both the mutable data
```
nbaksalyar: I think this comment should be replaced with something like
```suggestion
/// Implements functions for sequenced Mutable Data.
```

src/mutable_data.rs
```diff
@@ -133,3 +681,103 @@ impl MutableDataRef {
...
+
+impl UnseqEntryActions {
+    /// Insert a new key-value pair
+    pub fn ins(mut self, key: Vec<u8>, content: Vec<u8>) -> Self {
+        let _ = self.actions.insert(key, UnseqEntryAction::Ins(content));
+        self
+    }
+
+    /// Update existing key-value pair
+    pub fn update_unseq(mut self, key: Vec<u8>, content: Vec<u8>) -> Self {
```
nbaksalyar: Probably it'd be better to have a uniform API here?
I'd suggest using just `update` here, but ultimately I'd leave this to your judgement.

src/mutable_data.rs
```diff
@@ -133,3 +681,103 @@ impl MutableDataRef {
...
+        let _ = self.actions.insert(key, UnseqEntryAction::Update(content));
+        self
+    }
+
+    /// Delete existing key
+    pub fn del_unseq(mut self, key: Vec<u8>) -> Self {
```
nbaksalyar: Same as above, perhaps `del` would be better here?

src/mutable_data.rs
lionel-faber: Yea. That was my understanding too. This is meant to be just a temporary placeholder until we properly implement BLS signature verification

PR #10, opened 2019-06-04
Introduce MData enum and add more RPCs

lionel1704:
- Introduce MutableData enum to group sequenced and unsequenced types
- Add variants to Request and Reseponse to handle more RPCs for MData
operations

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -49,7 +49,14 @@ pub enum Request {
     GetUnpubIData(XorName),
     PutUnpubIData(UnpubImmutableData),
     DeleteUnpubIData(XorName),
-
+    DeleteMData {
+        // Address of the mutable data to be fetched
+        address: MutableDataRef,
+        // Requester public key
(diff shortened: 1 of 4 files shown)


lionel-faber:
Closing and reopening to trigger Travis CI

src/request.rs
```diff
@@ -49,7 +49,14 @@ pub enum Request {
     GetUnpubIData(XorName),
     PutUnpubIData(UnpubImmutableData),
     DeleteUnpubIData(XorName),
-
+    DeleteMData {
+        // Address of the mutable data to be fetched
+        address: MutableDataRef,
```
nbaksalyar, line 54: I think we might have to add the specification of `MutableDataKind` (as in `unsequenced` or `sequenced`) to disambiguate it further, but since both types exist in the same name space, I can't see any immediate disadvantages to having the same RPC for both without having to specify the actual type. So I'll leave it as a comment and we can discuss this later.

PR #14, opened 2019-06-04
Implement the basic functions of AppendOnlyData

nbaksalyar:
This PR implements some of the `AppendOnlyData` functions as they are described in the RFC.
Currently, it lacks the permissions handling and response types.

### DIFF
--- a/src/appendable_data.rs
+++ b/src/appendable_data.rs
@@ -49,7 +49,8 @@ pub struct UnpubPermissionSet {
 }
 
 impl UnpubPermissionSet {
-    fn is_allowed(&self, action: Action) -> bool {
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    pub fn is_allowed(&self, action: Action) -> bool {
         match action {
             Action::Read => self.read,
             Action::Append => self.append,
@@ -65,7 +66,8 @@ pub struct PubPermissionSet {
 }
 
 impl PubPermissionSet {
-    fn is_allowed(&self, action: Action) -> Option<bool> {
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    pub fn is_allowed(&self, action: Action) -> Option<bool> {
         match action {
             Action::Read => Some(true), // It's published data, so it's always allowed to read it.
             Action::Append => self.append,
--- a/src/appendable_data.rs
+++ b/src/appendable_data.rs
@@ -1,10 +1,11 @@
 // Copyright 2019 MaidSafe.net limited.
 //
-// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
-// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
-// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-// KIND, either express or implied. Please review the Licences for the specific language governing
-// permissions and limitations relating to use of the SAFE Network Software.
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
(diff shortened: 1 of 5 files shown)


src/errors.rs
```diff
@@ -9,11 +9,11 @@
 
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
-use std::error::Error;
+use std::error::Error as StdError;
```
mrcnski: Minor style point: I would delete this line and just refer to it as `::std::error::Error`, since I personally don't like renaming imports unless they are heavily used.

src/appendable_data.rs
```diff
@@ -0,0 +1,542 @@
...
+        match action {
+            Action::Read => Some(true), // It's published data, so it's always allowed to read it.
+            Action::Append => self.append,
+            Action::ManagePermissions => self.manage_permissions,
+        }
+    }
+}
+
+#[derive(Copy, Debug, Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+pub struct AppendOnlyDataRef {
```
mrcnski, line 81: Looks like this struct is identical to `MutableDataRef`. Are separate structs necessary here?

src/appendable_data.rs
```diff
@@ -0,0 +1,542 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
```
lionel-faber: License needs to be changed :smile:

src/appendable_data.rs
```diff
@@ -0,0 +1,542 @@
...
+/// Common methods for all `AppendOnlyData` flavours.
+pub trait AppendOnlyData<P> {
+    // /// Get a list of permissions for the provided user from the last entry in the permissions list.
+    // fn user_permissions(&self, user: &User) -> Result<&PubPermissionSet, ClientError>;
+
+    /// Return a value for the given key (if it is present).
+    fn get(&self, key: &[u8]) -> Option<&Vec<u8>>;
+
+    /// Get a list of keys and values with the given indexes.
+    fn at(&self, start: Index, end: Index) -> Option<&[(Vec<u8>, Vec<u8>)]>;
```
lionel-faber: Should we consider changing the name of this function?
`at` seems like getting an entry _at_ a particular index. Maybe we can consider something like `in_range` or `between` or `from_range`

src/appendable_data.rs
```diff
@@ -0,0 +1,542 @@
...
+    fn entry_index(&self) -> u64;
+
+    /// Return the last owners index.
+    fn owners_index(&self) -> u64;
+
+    /// Return the last permissions index.
+    fn permissions_index(&self) -> u64;
+
+    /// Get a complete list of permissions from the entry in the permissions list at the specified index.
+    fn permissions(&self, start: Index, end: Index) -> Option<&[P]>;
```
lionel-faber: We can perhaps consider including `range` in this function and in the `owners()` function too

src/appendable_data.rs
nbaksalyar: Agreed, and I also just spotted "indexes" which should be spelled "indices" :smile: Thanks!

src/appendable_data.rs
nbaksalyar: I'd be in favour of separating concerns here. They may look identical, but semantically they refer to different things: then, if the `AppendOnlyData` addressing schema changes (like e.g. we replace `type_tag` with something else), we wouldn't need to decouple it from `MutableDataRef`.

src/errors.rs
Yoga07: I guess, without renaming, it'd conflict with the enum `Error`?

mrcnski: If you refer to it using its entire name (`::std::error::Error`) it shouldn't conflict with anything.

PR #17, opened 2019-06-06
refactor/idata: calculate name when creating new unpublished idata

m-cat:
Fixes https://github.com/maidsafe/safe_client_libs/issues/828


src/immutable_data.rs
```diff
@@ -121,26 +124,17 @@ mod tests {
...
-        assert_eq!(idata3.name(), idata3.name());
-
-        assert_ne!(idata1.name(), idata2.name());
-        assert_ne!(idata1.name(), idata3.name());
-        assert_ne!(idata2.name(), idata3.name());
+        let idata1 = UnpubImmutableData::new(data1.clone(), owner1);
+        let idata2 = UnpubImmutableData::new(data1, owner2);
+        let idata3 = UnpubImmutableData::new(data2, owner1);
+
+        assert_eq!(idata1.name, idata1.name);
```
nbaksalyar: This would test for `Eq`/`PartialEq` implementation of `XorName`, so I think we shouldn't really need this?
The `assert_ne!` block should be enough.

PR #21, opened 2019-06-10
Properly handle mutable data permissions

lionel1704:
closes maidsafe/safe_client_libs#827


src/mutable_data.rs
```diff
@@ -8,13 +8,15 @@
 // Software.
 
 use crate::errors::{EntryError, Error};
+use crate::request::{Request, Requester};
+use crate::PublicKey;
 use crate::XorName;
+use maidsafe_utilities::serialisation::serialise;
```
nbaksalyar: We should use `bincode::serialize` instead, so `maidsafe_utilities` dependency is not required.

src/lib.rs
```diff
@@ -61,7 +61,7 @@ mod errors;
 mod identity;
 mod immutable_data;
 pub mod mutable_data;
-mod public_key;
+pub mod public_key;
```
nbaksalyar: Why do we need to make this module public? The contents of it is re-exported as `pub use public_key::PublicKey`, so I'd probably just leave it as private.

src/mutable_data.rs
```diff
@@ -297,18 +309,67 @@ macro_rules! impl_mutable_data {
...
+            fn is_action_allowed(&self, requester: &PublicKey, action: Action) -> bool {
+                match self.permissions.get(requester) {
                     Some(perms) => perms.is_allowed(action),
                     None => false,
                 }
             }
         }
     };
 }
+fn check_permissions_for_key(permissions: &PermissionSet, request: Request) -> Result<(), Error> {
```
nbaksalyar, line 336: Well done, really neat implementation here. :slightly_smiling_face:

src/mutable_data.rs
```diff
@@ -50,6 +51,14 @@ impl Value {
     pub fn new(data: Vec<u8>, version: u64) -> Self {
         Value { data, version }
     }
+
+    pub fn data(&self) -> Vec<u8> {
+        self.data.clone()
```
nbaksalyar, line 56: Making `data` pub in struct would be preferable, but we'll make this change in one of the next PRs.

nbaksalyar:
Rebased & integrated in #26

PR #22, opened 2019-06-10
Add Safecoin definitions

nbaksalyar:
This PR adds RPCs and structures required for the Safecoin implementation.
Co-authored by @Fraser999 and @jeanphilippeD.

### DIFF
--- a/src/coins.rs
+++ b/src/coins.rs
@@ -7,22 +7,18 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
-use crate::errors::Error;
+use crate::errors::{Error, Result};
 use serde::{Deserialize, Serialize};
-use std::result;
 use std::{
     fmt::{self, Debug, Display, Formatter},
     str::FromStr,
 };
 
-/// A specialised `Result` type for safecoin.
-pub type Result<T> = result::Result<T, Error>;
+/// The conversion from coin to raw value
+const COIN_TO_RAW_POWER_OF_10_CONVERSION: u32 = 9;
 
 /// The conversion from coin to raw value
-pub const COIN_TO_RAW_POWER_OF_10_CONVERSION: u32 = 9;
-
-/// The conversion from coin to raw value
-pub const COIN_TO_RAW_CONVERSION: u64 = 1_000_000_000;
+const COIN_TO_RAW_CONVERSION: u64 = 1_000_000_000;
 
 /// The maximum amount of safecoin represented by a single `Coins`.
 pub const MAX_COINS_VALUE: Coins =
--- a/src/coins.rs
+++ b/src/coins.rs
@@ -1,0 +1,192 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+use crate::errors::Error;
+use serde::{Deserialize, Serialize};
+use std::result;
+use std::{
+    fmt::{self, Debug, Display, Formatter},
+    str::FromStr,
+};
+
+/// A specialised `Result` type for safecoin.
+pub type Result<T> = result::Result<T, Error>;
+
+/// The conversion from coin to raw value
+pub const COIN_TO_RAW_POWER_OF_10_CONVERSION: u32 = 9;
+
+/// The conversion from coin to raw value
+pub const COIN_TO_RAW_CONVERSION: u64 = 1_000_000_000;
+
+/// The maximum amount of safecoin represented by a single `Coins`.
+pub const MAX_COINS_VALUE: Coins =
+    Coins((u32::max_value() as u64 + 1) * COIN_TO_RAW_CONVERSION - 1);
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+/// Structure representing a safecoin amount.
+pub struct Coins(u64);
+
+impl Coins {
+    /// New value from a number of nano coin.
+    pub fn from_nano(value: u64) -> Result<Self> {
+        if value > MAX_COINS_VALUE.0 {
+            return Err(Error::ExcessiveValue);
+        }
+        Ok(Self(value))
+    }
+
+    /// The maximum value a `Coins` can represent.
+    pub fn max_value() -> Self {
+        MAX_COINS_VALUE
+    }
+
+    /// Total coin expressed in number of nano coin.
+    pub fn as_nano(self) -> u64 {
+        self.0
+    }
+
+    // Computes self + rhs, returning None if overflow occurred.
+    pub fn checked_add(self, rhs: Coins) -> Option<Coins> {
+        self.0
+            .checked_add(rhs.0)
+            .and_then(|nano| Coins::from_nano(nano).ok())
+    }
+
+    // Computes self - rhs, returning None if overflow occurred.
+    pub fn checked_sub(self, rhs: Coins) -> Option<Coins> {
+        self.0
+            .checked_sub(rhs.0)
+            .and_then(|nano| Coins::from_nano(nano).ok())
+    }
+}
+
+impl FromStr for Coins {
+    type Err = Error;
+
+    fn from_str(value_str: &str) -> Result<Self> {
+        let mut itr = value_str.splitn(2, '.');
+        let converted_units = {
+            let units = itr
+                .next()
+                .and_then(|s| s.parse::<u64>().ok())
+                .ok_or(Error::FailedToParse)?;
+
+            units
+                .checked_mul(COIN_TO_RAW_CONVERSION)
+                .ok_or_else(|| Error::ExcessiveValue)?
+        };
+
+        let remainder = {
+            let remainder_str = itr.next().unwrap_or_default().trim_end_matches('0');
+
+            if remainder_str.is_empty() {
+                0
+            } else {
+                let parsed_remainder = remainder_str
+                    .parse::<u64>()
+                    .map_err(|_| Error::FailedToParse)?;
+
+                let remainder_conversion = COIN_TO_RAW_POWER_OF_10_CONVERSION
+                    .checked_sub(remainder_str.len() as u32)
+                    .ok_or_else(|| Error::LossOfPrecision)?;
+                parsed_remainder * 10_u64.pow(remainder_conversion)
+            }
+        };
+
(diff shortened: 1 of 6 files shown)


src/request.rs
```diff
@@ -292,6 +293,51 @@ pub enum Request {
...
+        transaction_id: u64, // TODO: Use the trait UUID
+        message_id: MessageId,
+    },
+    /// Get current wallet balance
+    GetBalance {
+        coins_balance_id: XorName,
+        message_id: MessageId,
+    },
+
+    // --- Client (Owner) to MM ---
```
ustulation: Nothing called `MM` anymore

src/response.rs
```diff
@@ -89,6 +103,42 @@ pub enum Response {
...
+    GetTransaction {
+        res: Result<Transaction, Error>,
+        msg_id: MessageId,
+    },
+    GetBalance {
+        res: Result<Coins, Error>,
+        msg_id: MessageId,
+    },
+
+    // --- Client (Owner) to MM ---
```
ustulation: Nothing called `MM` anymore

src/coins.rs
```diff
@@ -0,0 +1,192 @@
...
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+use crate::errors::Error;
+use serde::{Deserialize, Serialize};
+use std::result;
```
Fraser999: This should be part of the `use std` group in the line following.

src/coins.rs
```diff
@@ -0,0 +1,192 @@
...
+use std::{
+    fmt::{self, Debug, Display, Formatter},
+    str::FromStr,
+};
+
+/// A specialised `Result` type for safecoin.
+pub type Result<T> = result::Result<T, Error>;
```
Fraser999: I don't think we should limit this alias to just coins.rs.

src/coins.rs
```diff
@@ -0,0 +1,192 @@
...
+
+/// The conversion from coin to raw value
+pub const COIN_TO_RAW_POWER_OF_10_CONVERSION: u32 = 9;
```
Fraser999: This probably shouldn't be `pub`.  Same comment for `COIN_TO_RAW_CONVERSION`.

src/coins.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Total coin expressed in number of nano coin.
+    pub fn as_nano(self) -> u64 {
+        self.0
+    }
+
+    // Computes self + rhs, returning None if overflow occurred.
+    pub fn checked_add(self, rhs: Coins) -> Option<Coins> {
+        self.0
+            .checked_add(rhs.0)
+            .and_then(|nano| Coins::from_nano(nano).ok())
```
Fraser999, line 54: I guess returning an `Option` is consistent with the built in types' `checked_xxx` operations, but it's a bit inconsistent with `Coins::from_nano()` and `Coins::from_str()` which return `Err(Error::ExcessiveValue)` for this case.

src/coins.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    // Computes self - rhs, returning None if overflow occurred.
+    pub fn checked_sub(self, rhs: Coins) -> Option<Coins> {
+        self.0
+            .checked_sub(rhs.0)
+            .and_then(|nano| Coins::from_nano(nano).ok())
```
Fraser999, line 61: As above, maybe returning a new error variant (e.g. `NegativeValue`) would give a more consistent API?

src/errors.rs
```diff
@@ -54,6 +54,17 @@ pub enum Error {
     /// Network error occurring at Vault level which has no bearing on clients, e.g. serialisation
     /// failure or database failure
     NetworkOther(String),
+    /// While parsing or converting, precision would be lost.
+    LossOfPrecision,
+    /// The safecoin amount would exceed
+    /// [the maximum value for `Coin`](constant.MAX_COINS_VALUE.html).
```
Fraser999: This link will be broken if we don't expose `MAX_COINS_VALUE` in lib.rs.

Also, `Coin` should be `Coins`.

src/errors.rs
```diff
@@ -54,6 +54,17 @@ pub enum Error {
...
+    ExcessiveValue,
+    /// Failed to parse the string as a `Coin`.
```
Fraser999: `Coin` should be `Coins`.  We could also maybe link to the docs:
```
/// Failed to parse the string as a valid [`Coins`](struct.Coins.html) value.
```

src/errors.rs
```diff
@@ -54,6 +54,17 @@ pub enum Error {
     /// Network error occurring at Vault level which has no bearing on clients, e.g. serialisation
     /// failure or database failure
     NetworkOther(String),
+    /// While parsing or converting, precision would be lost.
```
Fraser999: This only applies to parsing now.

src/errors.rs
```diff
@@ -95,6 +106,16 @@ impl Display for Error {
             }
             Error::InvalidSignature => write!(f, "Failed signature validation"),
             Error::NetworkOther(ref error) => write!(f, "Error on Vault network: {}", error),
+            Error::LossOfPrecision => {
+                write!(f, "Lost precision on the number of coins during conversion")
```
Fraser999: Should be "parsing" rather than `conversion".  Same comment for `description()` variant.

src/lib.rs
```diff
@@ -65,6 +66,7 @@ mod public_key;
 pub mod request;
 pub mod response;
 
+pub use coins::Coins;
```
Fraser999: I think we should expose `MAX_COINS_VALUE` too.

src/lib.rs
```diff
@@ -74,10 +76,19 @@ pub use identity::{
...
+    distributions::{Distribution, Standard},
+    Rng,
+};
 use serde::{Deserialize, Serialize};
+pub use sha3::Sha3_512 as Ed25519Digest;
 use std::fmt::{self, Debug, Display, Formatter};
 
-pub use sha3::Sha3_512 as Ed25519Digest;
+/// Permissions for an app stored by the Elders.
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize, Default)]
```
Fraser999: Probably should be `Copy` too?

src/lib.rs
```diff
@@ -115,6 +136,13 @@ impl Display for XorName {
 #[derive(Ord, PartialOrd, Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize, Hash)]
 pub struct MessageId(pub XorName);
 
+impl MessageId {
+    /// Generate a new `MessageId` with random content.
+    pub fn new() -> MessageId {
```
mrcnski, line 169: I wonder if `new` is the best name for this function. Maybe `random` or `gen_random`?

src/lib.rs
```diff
@@ -107,6 +118,16 @@ impl Display for XorName {
     }
 }
 
+impl Distribution<XorName> for Standard {
```
Fraser999, line 127: Should this be a test-only impl?

src/request.rs
```diff
@@ -42,6 +43,15 @@ pub struct AppendOperation {
     requester: Requester,
```
mrcnski: I think `requester` should be removed here.

src/lib.rs
```diff
@@ -107,6 +118,16 @@ impl Display for XorName {
...
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> XorName {
+        let mut ret = [0u8; XOR_NAME_LEN];
```
Fraser999: I think we can just do `XorName(rng.gen())` here rather than iterating each `u8`.

src/request.rs
```diff
@@ -42,6 +43,15 @@ pub struct AppendOperation {
...
 }
 
+/// Wrapper message that contains the request, message ID, and the requester ID,
+/// which should also contain the signature(s) if it's sent by the owner(s).
```
Fraser999: "signature(s)" → "signature" since we won't aggregate the sigs - that'll be done before sending the request.

src/request.rs
```diff
@@ -42,6 +43,15 @@ pub struct AppendOperation {
...
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+pub struct Message {
```
Fraser999: Should we try and convey that this is a wrapper for Client requests?  Not sure of the best name to do that, but `Message` implies requests and responses to me.

src/response.rs
```diff
@@ -7,13 +7,27 @@
...
 use crate::immutable_data::UnpubImmutableData;
 use crate::mutable_data::{SeqMutableData, UnseqMutableData, Value};
-use crate::MessageId;
+use crate::{AppPermissions, MessageId};
+use rust_sodium::crypto::sign;
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, BTreeSet};
 
+/// Safecoin transaction.
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize, Debug)]
```
Fraser999: Maybe `Copy` too?

src/coins.rs
mrcnski: Should this be in `errors.rs` or `lib.rs`?

src/errors.rs
```diff
@@ -54,6 +54,17 @@ pub enum Error {
...
     NetworkOther(String),
+    /// While parsing or converting, precision would be lost.
+    LossOfPrecision,
+    /// The safecoin amount would exceed
+    /// [the maximum value for `Coin`](constant.MAX_COINS_VALUE.html).
+    ExcessiveValue,
+    /// Failed to parse the string as a `Coin`.
+    FailedToParse,
+    /// Transaction ID already exists.
+    TransactionIdExists,
```
mrcnski, line 71: Maybe we could store the Id in this variant?

src/lib.rs
mrcnski: Can we separate the `pub use` from the `use` statements? It's hard for me to tell what is being publicly exported.

src/errors.rs
```diff
@@ -54,6 +54,17 @@ pub enum Error {
     /// Network error occurring at Vault level which has no bearing on clients, e.g. serialisation
     /// failure or database failure
     NetworkOther(String),
+    /// While parsing or converting, precision would be lost.
+    LossOfPrecision,
+    /// The safecoin amount would exceed
+    /// [the maximum value for `Coin`](constant.MAX_COINS_VALUE.html).
+    ExcessiveValue,
+    /// Failed to parse the string as a `Coin`.
+    FailedToParse,
```
mrcnski, line 69: Maybe store the bad string in this variant?

src/lib.rs
mrcnski: Raised https://github.com/maidsafe/safe-nd/pull/23

PR #30, opened 2019-06-14
Many changes to move away from old temporary code

Fraser999:


src/append_only_data.rs
```diff
@@ -282,27 +296,26 @@ impl Permissions for PubPermissions {
...
 
 #[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash)]
 struct AppendOnly<P: Permissions> {
-    name: XorName,
-    tag: u64,
+    address: Address,
     data: Vec<(Vec<u8>, Vec<u8>)>,
     permissions: Vec<P>,
-    owners: Vec<Owners>,
+    owners: Vec<Owner>,
```
nbaksalyar, line 306: Maybe add a comment here stating that this is a history of ownership rather than a list of owners? Vec< Owner > doesn’t convey that I think

PR #31, opened 2019-06-17
Add challenge and change signature

octol:


src/lib.rs
```diff
@@ -152,7 +152,7 @@ pub enum Message {
     Request {
         request: Request,
         message_id: MessageId,
-        requester: Requester,
+        signature: Option<Signature>,
```
nbaksalyar, line 155: Could you please also remove the `Requester` enum altogether @octol, now that we don't need it?

nbaksalyar: This will be handled in a separate PR as a separate task. See https://github.com/maidsafe/safe_client_libs/issues/848

PR #32, opened 2019-06-19
Remove Requester and overhaul perms handling

nbaksalyar:
Requester has been made redundant with the change in the way we approach permissions handling client/app keys. This PR also exposes some of the necessary symbols from internal modules and fixes the `threshold_crypto` dependency.

### DIFF
--- a/src/append_only_data.rs
+++ b/src/append_only_data.rs
@@ -7,7 +7,7 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
-use crate::{Error, PublicKey, Requester, Result, XorName};
+use crate::{Error, PublicKey, Result, XorName};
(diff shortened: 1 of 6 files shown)


src/append_only_data.rs
```diff
@@ -206,22 +206,14 @@ impl Permissions for UnpubPermissions {
...
-                    }
-                } else {
-                    false
-                }
-            }
-            Requester::Owner(_sig) => {
-                // Temp
-                true
+    fn check_permission(&self, requester: PublicKey, action: Action) -> bool {
+        if self.permissions.contains_key(&requester) {
```
Fraser999: Do we need the `if` statement?  Can't we just go straight for `get()`?

nbaksalyar: Good point, thanks :)

PR #33, opened 2019-06-19
feat/adata: implement shell and owner_at_index getter functions for AppendOnly Data and minor fixes

Yoga07:


src/append_only_data.rs
```diff
@@ -547,6 +568,57 @@ impl SeqAppendOnlyData<PubPermissions> {
             },
         }
     }
+
+    pub fn shell(&self, index: u64) -> Result<Self> {
+
+        if self.entry_index() < index {
```
nbaksalyar: I would suggest replacing this with
```suggestion
        if index > self.entry_index() {
```
to make the intent more clear (checking for overflows).

src/append_only_data.rs
```diff
@@ -547,6 +568,57 @@ impl SeqAppendOnlyData<PubPermissions> {
             },
         }
     }
+
+    pub fn shell(&self, index: u64) -> Result<Self> {
+
```
nbaksalyar: There's extraneous whitespace, could you please run rustfmt on it?

Yoga07:
> I wonder if we can avoid code duplication on this.
> Would it be possible to move the `shell` impl into the `impl_appendable_data` macro? I think the shell should be mostly the same for all types of `AppendOnlyData`.

Ah I did try writing the function inside the trait. It required the `Sized` trait to be implemented to the structs to return a `Self`.

nbaksalyar:
>I did try writing the function inside the trait. It required the Sized trait to be implemented to the structs to return a Self.

It does not necessarily have to be defined in the trait @Yoga07.
So what I mean is we have this macro `impl_appendable_data` which implements trait methods for different AppendOnlyData types, but it can also be used to implement struct-level methods.

So it can be just this, after the `pub struct $flavour<P>` block inside the macro:

```rust
impl<P> $flavour<P>
where
  P: Permissions + std::hash::Hash + Clone,
{
  pub fn shell(&self, index: u64) -> Result<Self> {
    ..
  }
}
```

src/lib.rs
```diff
@@ -66,7 +66,7 @@ mod request;
 mod response;
 
 pub use append_only_data::{
-    Address as ADataAddress, AppendOnlyData, Index as ADataIndex, Indices as ADataIndices,
+    Action, Address as ADataAddress, AppendOnlyData, Index as ADataIndex, Indices as ADataIndices,
```
nbaksalyar: I think this should be
```suggestion
    Action as ADataAction, Address as ADataAddress, AppendOnlyData, Index as ADataIndex, Indices as ADataIndices,
```
to be not confused with `MDataAction`.

src/append_only_data.rs
```diff
@@ -265,11 +265,21 @@ impl Permissions for PubPermissions {
...
+    pub public_key: PublicKey,
     /// The current index of the data when this ownership change happened
-    data_index: u64,
+    pub data_index: u64,
     /// The current index of the permissions when this ownership change happened
-    permissions_index: u64,
+    pub permissions_index: u64,
+}
+
+impl Owner {
```
Fraser999: Is there a need for this impl since the members are public now?

src/append_only_data.rs
```diff
@@ -357,14 +370,49 @@ macro_rules! impl_appendable_data {
...
+                    if owner.data_index() <= index {
+                        owner_vec.push(owner.clone());
+                    }
+                }
+
+                Ok(Self {
+                    inner: AppendOnly {
+                        address: self.inner.address,
+                        data: Vec::new(),
+                        permissions: perm_vec.clone(),
```
Fraser999: I don't think we need to clone here or on the line below.

src/append_only_data.rs
```diff
@@ -417,6 +465,14 @@ macro_rules! impl_appendable_data {
                 }
             }
 
+            fn fetch_owner_at_index(&self, owners_index: u64) -> Option<&Owner> {
+                if self.inner.owners.len() >= owners_index as usize {
```
Fraser999: This could just be `self.inner.owners.get(owners_index as usize)`

src/append_only_data.rs
```diff
@@ -357,14 +370,49 @@ macro_rules! impl_appendable_data {
...
+        impl<P> $flavour<P>
+        where
+            P: Permissions + std::hash::Hash + Clone,
+        {
+            pub fn shell(&self, index: u64) -> Result<Self> {
+                if index > self.entry_index() {
+                    return Err(Error::NoSuchEntry);
+                }
+
+                let mut perm_vec: Vec<P> = Vec::new();
```
Fraser999: We could probably just use `filter` here?  Or `take_while` if the indices are always held in increasing order?  E.g.
```rust
let permissions = self
    .inner
    .permissions
    .iter()
    .filter(|perm| perm.data_index() <= index)
    .cloned()
    .collect();
let owners = self
    .inner
    .owners
    .iter()
    .filter(|owner| owner.data_index <= index)
    .cloned()
    .collect();
```

src/append_only_data.rs
```diff
@@ -357,14 +360,50 @@ macro_rules! impl_appendable_data {
...
+                    .iter()
+                    .filter(|owner| owner.data_index <= index)
+                    .cloned()
+                    .collect();
+
+                Ok(Self {
+                    inner: AppendOnly {
+                        address: self.inner.address,
+                        data: Vec::new(),
+                        permissions: permissions,
```
nbaksalyar: Can be just
```suggestion
                        permissions,
```

src/append_only_data.rs
```diff
@@ -357,14 +360,50 @@ macro_rules! impl_appendable_data {
...
+                        owners: owners,
```
nbaksalyar: same here
```suggestion
                        owners,
```

PR #35, opened 2019-06-20
feat/public_id: serialise to and from base64

octol:


src/identity/client.rs
```diff
@@ -141,8 +147,18 @@ impl Debug for PublicId {
...
+        let pub_id = unwrap!(bincode::serialize(&self));
+        write!(formatter, "{}", base64::encode(&pub_id))
+    }
+}
+
+impl FromStr for PublicId {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let pub_id = base64::decode(s).map_err(|e| Error::NetworkOther(e.to_string()))?;
```
octol: I wanted to make this conversion automatic by implementing the necessary traits but wasn't successful yet

src/identity/mod.rs
```diff
@@ -71,3 +74,25 @@ struct BlsKeypairShare {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::str::FromStr;
+    use unwrap::unwrap;
+
+    #[test]
+    fn serialize_to_base64() {
```
mrcnski: Looks like a test for `app::FullId` is missing here.

src/identity/app.rs
```diff
@@ -99,6 +100,19 @@ impl PublicId {
...
+
+    /// Returns the PublicId serialised and encoded in base64.
+    pub fn encode_to_base64(&self) -> String {
+        let pub_id = unwrap!(bincode::serialize(&self));
+        base64::encode(&pub_id)
+    }
+
+    /// Create from base64 encoded string.
+    pub fn decode_from_base64(encoded: &str) -> Result<Self, Error> {
+        let pub_id = base64::decode(encoded).map_err(|e| Error::NetworkOther(e.to_string()))?;
```
Fraser999: Hmm - I'm not sure about this one.  The docs for `Error::NetworkOther` state "Network error occurring at Vault level which has no bearing on clients, e.g. serialisation failure or database failure".  However, I don't think vaults will normally be using this function for anything other than decoding a string which the user provided (e.g. via the config file).

Under those circumstances, it seems like we should maybe change `FailedToParse` to internally hold a String, and use that here and on the line below?

src/identity/mod.rs
```diff
@@ -71,3 +71,52 @@ struct BlsKeypairShare {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::ClientFullId;
+    use unwrap::unwrap;
+
+    #[test]
+    fn serialize_client_public_id_to_base64() {
```
Fraser999: I think the three tests' names should include "encode" rather than "serialize".

Also, I think we should probably test failure paths of decoding too.

src/identity/app.rs
```diff
@@ -99,6 +100,19 @@ impl PublicId {
...
     }
+
+    /// Returns the PublicId serialised and encoded in base64.
+    pub fn encode_to_base64(&self) -> String {
+        let pub_id = unwrap!(bincode::serialize(&self));
+        base64::encode(&pub_id)
+    }
+
+    /// Create from base64 encoded string.
+    pub fn decode_from_base64(encoded: &str) -> Result<Self, Error> {
```
Fraser999: I think we should make the method's signature:
```rust
pub fn decode_from_base64<T: ?Sized + AsRef<[u8]>>(encoded: &T) -> 
```
since this matches `base64::decode`'s input and is a bit more user-friendly.

src/identity/app.rs
```diff
@@ -99,6 +100,19 @@ impl PublicId {
     pub fn owner(&self) -> &ClientPublicId {
         &self.owner
     }
+
+    /// Returns the PublicId serialised and encoded in base64.
```
Fraser999: For all six doc comments, we should mention that we use standard base64.

src/identity/app.rs
octol: What about having `FailedToParseIdentity` in addition to `FailedToParse`?

Fraser999: Well, that would be better than using `NetworkOther` imo.  I assume it'd wrap a String?  If you were going this way, I'd recommend renaming `FailedToParse` to `FailedToParseCoins` though.

PR #36, opened 2019-06-21
Add functions for identities and refactor mutable data permission handling

lionel1704:
- Add functions to create App and Client identities with the provided
keys
- Verify signatures for both owner and application requests

### DIFF
--- a/src/identity/app.rs
+++ b/src/identity/app.rs
@@ -12,7 +12,7 @@ use crate::{ClientFullId, ClientPublicId, Ed25519Digest, PublicKey, Signature, X
 use rand::{CryptoRng, Rng};
 use serde::{Deserialize, Serialize};
 use std::fmt::{self, Debug, Display, Formatter};
-use threshold_crypto::SecretKeyShare as BlsSecretKeyShare;
+use threshold_crypto::{SecretKey as BlsSecretKey, SecretKeyShare as BlsSecretKeyShare};
 
 /// A struct holding a keypair variant and the corresponding public ID for a network App.
 #[derive(Serialize, Deserialize)]
@@ -60,6 +60,17 @@ impl FullId {
     /// Returns the public ID.
     pub fn public_id(&self) -> &PublicId {
         &self.public_id
+    }
+
+    // TODO: Remove this once the authenticator is updated
+    // to create random FullIds instead of AppKeys / ClientKeys
+    /// Constructs a `FullId` for a given BLS secret key.
+    #[doc(hidden)]
+    pub fn with_keys(bls_sk: BlsSecretKey, owner: PublicKey) -> Self {
+        Self::new(
+            ClientFullId::with_bls_key(bls_sk),
+            ClientPublicId::new(XorName::from(owner), owner),
+        )
     }
 }
 
--- a/src/identity/client.rs
+++ b/src/identity/client.rs
@@ -96,6 +96,27 @@ impl FullId {
     pub fn public_id(&self) -> &PublicId {
         &self.public_id
     }
+
+    // TODO: Remove this once the authenticator is updated
+    // to create random FullIds instead of AppKeys / ClientKeys
+    /// Constructs a `FullId` with a particular BLS secret key.
+    #[doc(hidden)]
+    pub fn with_bls_key(bls_sk: BlsSecretKey) -> Self {
+        let bls_pk = bls_sk.public_key();
+        let bls_keypair = BlsKeypair {
+            secret: SerdeSecret(bls_sk),
+            public: bls_pk,
+        };
+        let public_key = PublicKey::Bls(bls_pk);
+        let public_id = PublicId {
+            name: public_key.into(),
+            public_key,
+        };
+        Self {
+            keypair: Keypair::Bls(bls_keypair),
+            public_id,
+        }
+    }
 }
 
 /// A struct representing the public identity of a network Client.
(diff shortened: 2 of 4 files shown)


src/mutable_data.rs
```diff
@@ -214,11 +220,22 @@ macro_rules! impl_mutable_data {
...
 
-            fn check_permissions(&self, request: Request, requester: PublicKey) -> Result<()> {
+            fn check_permissions(
+                &self,
+                request: Request,
+                msg_id: MessageId,
+                requester: PublicKey,
+                signature: Signature,
+            ) -> Result<()> {
+                verify_signature(signature, requester, request.clone(), msg_id)?;
```
nbaksalyar: I think this is better done outside of the `check_permissions` function.
So now it does 2 things while it can do just 1 and don't really care about the signatures.
What I'd propose is to leave `verify_signature` as it is, and remove the signature checking and the `signature` argument here.

src/mutable_data.rs
```diff
@@ -543,10 +570,10 @@ impl SeqMutableData {
             },
         );
 
-        if *self.owners() == requester
-            && (!insert.is_empty() && !self.is_action_allowed(&requester, Action::Insert))
-            || (!update.is_empty() && !self.is_action_allowed(&requester, Action::Update))
-            || (!delete.is_empty() && !self.is_action_allowed(&requester, Action::Delete))
+        if *self.owners() != requester
```
nbaksalyar, line 550: Nice catch :)

src/identity/client.rs
```diff
@@ -96,6 +96,24 @@ impl FullId {
     pub fn public_id(&self) -> &PublicId {
         &self.public_id
     }
+
+    /// Constructs a `FullId` with a particular BLS secret key.
+    pub fn with_bls_key(bls_sk: BlsSecretKey) -> Self {
```
Fraser999, line 104: To avoid code duplication, we should change `new_bls` to use this c'tor:
```rust
pub fn new_bls<T: CryptoRng + Rng>(rng: &mut T) -> Self {
    let bls_secret_key = rng.gen::<BlsSecretKey>();
    Self::with_bls_key(bls_secret_key)
}
```

src/identity/client.rs
```diff
@@ -118,6 +136,10 @@ impl PublicId {
     pub fn public_key(&self) -> &PublicKey {
         &self.public_key
     }
+
+    pub fn new(name: XorName, public_key: PublicKey) -> Self {
```
Fraser999, line 146: Is this something needed only for testing?  Maybe we should feature gate it?

src/identity/mod.rs
```diff
@@ -32,6 +32,14 @@ pub enum PublicId {
     App(app::PublicId),
 }
 
+/// An enum representing the Full Id variants for a Client or App
+pub enum FullId {
```
Fraser999: Do we need this in safe-nd?  I don't think it's something vaults will ever need.  I'm not really able to even see a use case for it in SCL, but I'm not too familiar with that crate.

src/identity/app.rs
```diff
@@ -61,6 +61,14 @@ impl FullId {
     pub fn public_id(&self) -> &PublicId {
         &self.public_id
     }
+
+    /// Constructs a `FullId` for a given BLS secret key.
+    pub fn with_keys(bls_sk: BlsSecretKey, owner: PublicKey) -> Self {
```
Fraser999, line 69: Is this c'tor only required for testing?  Same question for `ClientFullId::with_bls_key()`.

src/lib.rs
```diff
@@ -78,7 +78,7 @@ pub use identity::{
     app::{FullId as AppFullId, PublicId as AppPublicId},
     client::{FullId as ClientFullId, PublicId as ClientPublicId},
     node::{FullId as NodeFullId, PublicId as NodePublicId},
-    PublicId,
+    FullId as FullIdentity, PublicId,
```
Fraser999: If we do keep the `FullId` enum, then can we just export it as `FullId` rather than `FullIdentity` just for consistency?

src/identity/mod.rs
lionel-faber: @Fraser999 I needed this wrapper so that I can add it to the mock [routing struct here](https://github.com/lionel1704/safe_client_libs/blob/78c2c063d69eacff185d4e1a685a5fc2174bc6ae/safe_core/src/client/mock/routing.rs#L86). Maybe once mock routing is deprecated we wouldn't need this.

src/identity/app.rs
lionel-faber: I needed them so that I could construct respective `FullId`s for already existing clients and apps. This is then assigned to the mock routing struct as I mentioned before.

src/identity/client.rs
lionel-faber: I needed it for the `FullId` constructors. Usage is [here](https://github.com/maidsafe/safe-nd/pull/36/files/b5c72383ce2b746280bcda0a884144e77f70d3d5#diff-c0b1a46ed40f1a21040e004922465426R69)

src/identity/client.rs
lionel-faber: I needed to create a `ClientFullId` given a specific BLS key that was created during account creation.

src/lib.rs
```diff
@@ -198,3 +198,14 @@ pub enum Challenge {
     Request(Vec<u8>),
     Response(PublicId, Signature),
 }
+
+/// Verify that a signature is valid for a given Request + MessageId combination
+pub fn verify_signature(
+    signature: Signature,
```
Fraser999: I think these arguments should all be refs - especially `request`.

src/identity/client.rs
```diff
@@ -118,6 +139,12 @@ impl PublicId {
     pub fn public_key(&self) -> &PublicKey {
         &self.public_key
     }
+
+    // TODO: Remove this once the authenticator is updated
+    // to create random FullIds instead of AppKeys / ClientKeys
+    pub fn new(name: XorName, public_key: PublicKey) -> Self {
```
nbaksalyar, line 146: This one should be `#[doc(hidden)]` too I suppose?

PR #37, opened 2019-06-22
fix/adata: fix permissions checking for AppendOnly Data and impl getter functions in AppendOnly Data Trait

Yoga07:
- Fixes requester permissions checking for  AppendOnly Data
- Additions and refactors to Requests and Responses
- Implements getter functions in `AppendOnly` trait


src/append_only_data.rs
```diff
@@ -337,7 +322,13 @@ pub trait AppendOnlyData<P> {
     /// Add a new owner entry.
     fn append_owner(&mut self, owner: Owner) -> Result<()>;
 
-    fn verify_requester(&self, requester: PublicKey, action: Action) -> Result<bool>;
+    /// Verifies permission for Non-Owners.
+    fn check_permissions_for_key(
+        &self,
+        requester: PublicKey,
+        permissions: &P,
+        request: Request,
```
nbaksalyar: This should be `&Request` to prevent unnecessary copying.

src/append_only_data.rs
```diff
@@ -399,6 +390,18 @@ macro_rules! impl_appendable_data {
                     },
                 })
             }
+
+            pub fn check_permission(&self, request: Request, requester: PublicKey) -> Result<()> {
+                if unwrap!(self.inner.owners.last()).public_key == requester {
```
nbaksalyar: Generally, we should not use `unwrap`s in the production code.
Given this function returns a `Result`, we can do something like this instead:
```suggestion
                if self.inner.owners.last().ok_or_else(|| Error::InvalidState)?.public_key == requester {
```

src/append_only_data.rs
```diff
@@ -429,10 +432,46 @@ macro_rules! impl_appendable_data {
...
+                    Request::GetAData(..)
+                    | Request::GetADataShell { .. }
+                    | Request::GetADataRange { .. }
+                    | Request::GetADataIndices(..)
+                    | Request::GetADataLastEntry(..)
+                    | Request::GetADataPermissions { .. }
+                    | Request::GetPubADataUserPermissions { .. }
+                    | Request::GetUnpubADataUserPermissions { .. }
+                    | Request::GetADataOwners { .. } => {
+                        if permissions.is_action_allowed(requester, Action::Read) {
```
nbaksalyar: `Action::Read` should be handled differently for published and unpublished data.
It should always return `true` for all read actions for published `AppendOnlyData`, so I would suggest either moving this function to the `Permissions` trait, or always return `true` for `Action::Read` in `PubPermissions::is_action_allowed`.

src/append_only_data.rs
```diff
@@ -429,10 +432,46 @@ macro_rules! impl_appendable_data {
...
+                            Ok(())
+                        } else {
+                            Err(Error::AccessDenied)
+                        }
+                    }
+                    Request::AddPubADataPermissions { .. }
+                    | Request::AddUnpubADataPermissions { .. }
+                    | Request::SetADataOwner { .. } => {
```
nbaksalyar: `Request::SetADataOwner` is available only to an existing owner, so it should be always `Error::AccessDenied` for non-owner requests.

src/append_only_data.rs
```diff
@@ -429,10 +432,46 @@ macro_rules! impl_appendable_data {
...
+                        if permissions.is_action_allowed(requester, Action::ManagePermissions) {
+                            Ok(())
+                        } else {
+                            Err(Error::AccessDenied)
+                        }
+                    }
+
+                    // Mutation permissions are checked later
```
nbaksalyar: Why? I don't think we make any distinction for update/delete/insert permissions as in the case of `MutableData`, so we can just check for `Action::Append` permissions here I guess.

src/append_only_data.rs
```diff
@@ -429,10 +432,46 @@ macro_rules! impl_appendable_data {
...
+                    Request::AppendSeq { .. } | Request::AppendUnseq { .. } => Ok(()),
+
+                    Request::DeleteAData { .. } => Err(Error::AccessDenied),
+
+                    _ => Err(Error::InvalidOperation),
```
Fraser999: Is it worth listing all remaining variants here rather than using `_` so that any future changes to the enum get caught here?

src/response.rs
```diff
@@ -153,11 +154,12 @@ impl fmt::Debug for Response {
                 GetUnpubADataUserPermissions(..) => "Response::GetUnpubADataUserPermissions",
                 AddUnpubADataPermissions(..) => "Response::AddUnpubADataPermissions",
                 AddPubADataPermissions(..) => "Response::AddPubADataPermissions",
-                AppendUnpubSeq(..) => "Response::AppendUnpubSeq",
-                AppendPubUnseq(..) => "Response::AppendPubUnseq",
-                AppendPubSeq(..) => "Response::AppendPubSeq",
-                AppendUnpubUnseq(..) => "Response::AppendUnpubUnseq",
+                AppendSeq(..) => "Response::AppendUnpubSeq",
```
Fraser999: Looks like the strings for this and `AppendUnseq` below didn't get updated.

src/append_only_data.rs
Yoga07: You mean only the AppendOnly Data variants? 🙂  It would be kinda long list if we include _Mutable_ and _Immutable Data_ variants here.

src/append_only_data.rs
```diff
@@ -246,6 +282,38 @@ impl Permissions for PubPermissions {
...
+        match request {
+            Request::GetAData(..)
+            | Request::GetADataShell { .. }
+            | Request::GetADataRange { .. }
+            | Request::GetADataIndices(..)
+            | Request::GetADataLastEntry(..)
+            | Request::GetADataPermissions { .. }
+            | Request::GetPubADataUserPermissions { .. }
+            | Request::GetADataOwners { .. } => Ok(()),
+            Request::AddPubADataPermissions { .. } | Request::SetADataOwner { .. } => {
```
nbaksalyar: > Request::SetADataOwner

This should be always forbidden for anyone but an owner.
Could you pls move it to the `Request::DeleteAData { .. } => Err(Error::AccessDenied),` block?

src/append_only_data.rs
```diff
@@ -205,6 +205,42 @@ impl Permissions for UnpubPermissions {
...
+            | Request::GetADataPermissions { .. }
+            | Request::GetUnpubADataUserPermissions { .. }
+            | Request::GetADataOwners { .. } => {
+                if self.is_action_allowed(requester, Action::Read) {
+                    Ok(())
+                } else {
+                    Err(Error::AccessDenied)
+                }
+            }
+            Request::AddUnpubADataPermissions { .. } | Request::SetADataOwner { .. } => {
```
nbaksalyar: Same for `SetADataOwner` here.

src/request.rs
```diff
@@ -42,6 +43,144 @@ impl AppendOnlyData {
...
+
+    pub fn get_owners(&self, idx: u64) -> Option<&ADataOwner> {
+        match self {
+            AppendOnlyData::PubSeq(data) => data.fetch_owner_at_index(idx),
+            AppendOnlyData::PubUnseq(data) => data.fetch_owner_at_index(idx),
+            AppendOnlyData::UnpubSeq(data) => data.fetch_owner_at_index(idx),
+            AppendOnlyData::UnpubUnseq(data) => data.fetch_owner_at_index(idx),
+        }
+    }
+    pub fn get_pub_user_permissions(
```
nbaksalyar, line 117: There seem to be some formatting problems - could you please run `rustfmt` on this file?

Yoga07: Hmm, I did run `rustfmt` @nbaksalyar. The CI passed too. 🤔

src/append_only_data.rs
```diff
@@ -206,10 +206,39 @@ impl Permissions for UnpubPermissions {
...
+        match request {
+            Request::GetAData(..)
+            | Request::GetADataShell { .. }
+            | Request::GetADataRange { .. }
+            | Request::GetADataIndices(..)
+            | Request::GetADataLastEntry(..)
+            | Request::GetADataPermissions { .. }
+            | Request::GetUnpubADataUserPermissions { .. }
+            | Request::GetADataOwners { .. } => {
+                if self.is_action_allowed(requester, Action::Read) {
```
Fraser999: Seems like it might be simpler if we changed `is_action_allowed()` to return a `Result<()>` too?

src/request.rs
nbaksalyar: Ah, could you please add a new line above this function definition to be consistent with the rest of the code style? :slightly_smiling_face:

src/append_only_data.rs
```diff
@@ -254,11 +283,35 @@ impl Permissions for PubPermissions {
...
+            | Request::GetPubADataUserPermissions { .. }
+            | Request::GetADataOwners { .. } => Ok(()),
+            Request::AddPubADataPermissions { .. } => {
+                if self.is_action_allowed(requester, Action::ManagePermissions) {
+                    Ok(())
+                } else {
+                    Err(Error::AccessDenied)
+                }
+            }
+            // Mutation permissions are checked later
```
nbaksalyar: This comment seems unnecessary now. Could you please remove it?

src/request.rs
```diff
@@ -42,6 +43,144 @@ impl AppendOnlyData {
...
+        &self,
+        user: ADataUser,
+        idx: u64,
+    ) -> Result<ADataPubPermissionSet, Error> {
+        match self {
```
Fraser999, line 122: This could maybe be simplified a bit:
```rust
match self {
    AppendOnlyData::PubSeq(data) => data.fetch_permissions_at_index(idx),
    AppendOnlyData::PubUnseq(data) => data.fetch_permissions_at_index(idx),
    _ => None,
}
.and_then(|permissions| permissions.permissions().get(&user))
.cloned()
.ok_or(Error::NoSuchEntry)
```

src/request.rs
```diff
@@ -42,6 +43,144 @@ impl AppendOnlyData {
...
+                    None => Err(Error::NoSuchEntry),
+                },
+                None => Err(Error::NoSuchEntry),
+            },
+            _ => Err(Error::NoSuchData),
+        }
+    }
+
+    pub fn get_shell(&self, idx: u64) -> Result<Self, Error> {
+        match self {
```
Fraser999, line 148: This could be:
```rust
use AppendOnlyData::*;
match self {
    PubSeq(adata) => adata.shell(idx).map(PubSeq),
    PubUnseq(adata) => adata.shell(idx).map(PubUnseq),
    UnpubSeq(adata) => adata.shell(idx).map(UnpubSeq),
    UnpubUnseq(adata) => adata.shell(idx).map(UnpubUnseq),
}
```

PR #42, opened 2019-06-25
refactor/response: remove TransferCoins response

m-cat:
We shouldn't have a response variant for TransferCoins: the request should be fire-and-forget and we should get the transaction status by sending a GetTransaction request.

Fixes https://github.com/maidsafe/safe_client_libs/issues/868


nbaksalyar:
This will be merged as part of #48. Thanks!

PR #43, opened 2019-06-25
Add CreateCoinBalance and PutAccount RPCs

nbaksalyar:


src/request.rs
```diff
@@ -308,6 +308,22 @@ pub enum Request {
     },
     /// Get current wallet balance
     GetBalance(XorName),
+    /// Create a new coin balance
+    CreateCoinBalance {
+        source: XorName,
```
Fraser999: This field should be removed, since the source will be defined by the sender of the message.

I just noticed too (although not added in this PR) that the same is true for `TransferCoins` and `GetBalance`.

Issue #44, opened 2019-06-26
Update encoding


Fraser999:
We need to make the following changes:

1. Merge the two error variants `FailedToParseCoins` and `FailedToParseIdentity(String)` into a single one `FailedToParse(String)` and ensure a meaningful message is added when constructing one of these variants (e.g. to replace the ones in coins.rs, consider something like `Error::FailedToParse("Can't parse coin units".to_string())` and `Error::FailedToParse("Can't parse coin remainder".to_string())`.
2. Create a utils.rs module.  It should contain:
    * the `verify_signature` method currently in lib.rs
    * a serialisation helper:
    ```rust
    pub(crate) fn serialise<T: Serialize>(data: &T) -> Vec<u8> {
        unwrap!(bincode::serialize(data))
    }
    ```
    * an encoding helper:
    ```rust
    pub(crate) fn encode<T: Serialize>(data: &T) -> String {
        let serialised = serialise(&data);
        multibase::encode(Base::Base32z, &serialised)
    }
    ```
    * a decoding helper:
    ```rust
    pub(crate) fn decode<I: AsRef<str>, O: DeserializeOwned>(encoded: &I) -> Result<O> {
        let (base, decoded) =
            multibase::decode(encoded).map_err(|e| Error::FailedToParse(e.to_string()))?;
        if base != Base::Base32z {
            return Err(Error::FailedToParse(format!(
                "Expected z-base-32 encoding, but got {:?}",
                base
            )));
        }
        Ok(bincode::deserialize(&decoded).map_err(|e| Error::FailedToParse(e.to_string()))?)
    }
    ```
3. Replace all occurrences of `bincode::serialize` with the new helper
4. Remove base64 from the crate and exclusively use multibase z-base-32 encoding
5. Change all existing helpers `encode_to_base64` to `encode_to_z_base_32` and similar change for decode helpers.
6. Provide encode/decode helpers (and tests) for
    * `XorName`
    * `PublicKey`
    * `AppPublicId`
    * `ClientPublicId`
    * `NodePublicId`
    * `PublicId`
    * `ADataAddress`
    * `IDataAddress`
    * `MDataAddress`

Yoga07:
Resolved in #45

PR #45, opened 2019-06-27
refactor/encode: introduces multibase z-base-32 encode/decode by swapping base64 encode/decode

Yoga07:
Resolves issue #44 

- Removed base64 crate from dependencies
- Added utils module containing wrappers for serialization and encode/decode
- Changed all occurrences of previous base64 encode/decode to the new z-base-32 encode/decode 
   wrappers
- Implemented encode/decode helper functions and tests for Xorname, PublicKey, 
  AppPublicId,ClientPublicId, NodePublicId, PublicId, ADataAddress, IDataAddress, MDataAddress

  This commit also includes a minor refactor for AppendOnly Data
- Moved AData impl from requests module to append_only_data module
- Introduced two macros to reduce LOC
- Moved permissions checking from types to AData enum
- Refactored return type for in_range function

### DIFF
--- a/src/append_only_data.rs
+++ b/src/append_only_data.rs
@@ -7,7 +7,8 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
-use crate::{Error, PublicKey, Request, Result, XorName};
+use crate::{utils, Error, PublicKey, Request, Result, XorName};
+use multibase::Decodable;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 
@@ -15,6 +16,7 @@ pub type PubSeqAppendOnlyData = SeqAppendOnlyData<PubPermissions>;
 pub type PubUnseqAppendOnlyData = UnseqAppendOnlyData<PubPermissions>;
 pub type UnpubSeqAppendOnlyData = SeqAppendOnlyData<UnpubPermissions>;
 pub type UnpubUnseqAppendOnlyData = UnseqAppendOnlyData<UnpubPermissions>;
+pub type Entries = Vec<(Vec<u8>, Vec<u8>)>;
 
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub enum User {
@@ -166,6 +168,16 @@ impl Address {
             | Address::UnpubUnseq { tag, .. } => *tag,
         }
     }
+
+    /// Returns the Address serialised and encoded in z-base-32.
+    pub fn encode_to_zbase32(&self) -> String {
+        utils::encode(&self)
+    }
+
+    /// Create from z-base-32 encoded string.
+    pub fn decode_from_zbase32<I: Decodable>(encoded: I) -> Result<Self> {
+        utils::decode(encoded)
+    }
 }
 
 pub trait Permissions {
@@ -315,7 +327,7 @@ pub struct Owner {
 #[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash)]
 struct AppendOnly<P: Permissions> {
     address: Address,
-    data: Vec<(Vec<u8>, Vec<u8>)>,
+    data: Entries,
     permissions: Vec<P>,
     // This is the history of owners, with each entry representing an owner.  Each single owner
     // could represent an individual user, or a group of users, depending on the `PublicKey` type.
@@ -331,13 +343,13 @@ pub trait AppendOnlyData<P> {
     fn get(&self, key: &[u8]) -> Option<&Vec<u8>>;
 
     /// Return the last entry in the Data (if it is present).
-    fn last(&self) -> Option<(Vec<u8>, Vec<u8>)>;
+    fn last_entry(&self) -> Option<(Vec<u8>, Vec<u8>)>;
 
     /// Get a list of keys and values with the given indices.
-    fn in_range(&self, start: Index, end: Index) -> Option<Vec<(Vec<u8>, Vec<u8>)>>;
+    fn in_range(&self, start: Index, end: Index) -> Option<Entries>;
 
     /// Return all entries.
-    fn entries(&self) -> &Vec<(Vec<u8>, Vec<u8>)>;
+    fn entries(&self) -> &Entries;
 
     /// Return the address of this AppendOnlyData.
     fn address(&self) -> &Address;
@@ -437,25 +449,6 @@ macro_rules! impl_appendable_data {
                     },
                 })
             }
-
-            pub fn check_permission(&self, request: &Request, requester: PublicKey) -> Result<()> {
-                if self
-                    .inner
-                    .owners
-                    .last()
-                    .ok_or_else(|| Error::NoSuchData)?
-                    .public_key
-                    == requester
-                {
-                    Ok(())
-                } else {
-                    self.inner
-                        .permissions
-                        .last()
-                        .ok_or_else(|| Error::NoSuchData)?
-                        .check_permissions_for_key(requester, request)
-                }
-            }
         }
 
         impl<P> AppendOnlyData<P> for $flavour<P>
@@ -493,7 +486,7 @@ macro_rules! impl_appendable_data {
                     .find_map(|(k, v)| if k.as_slice() == key { Some(v) } else { None })
             }
 
-            fn last(&self) -> Option<(Vec<u8>, Vec<u8>)> {
+            fn last_entry(&self) -> Option<(Vec<u8>, Vec<u8>)> {
                 match self.inner.data.last() {
                     Some(tup) => Some(tup.clone()),
                     None => None,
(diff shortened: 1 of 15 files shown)


src/request.rs
```diff
@@ -308,6 +162,22 @@ pub enum Request {
     },
     /// Get current wallet balance
     GetBalance(XorName),
+    /// Create a new coin balance
+    CreateCoinBalance {
+        source: XorName,
```
nbaksalyar: Could you please remove the `source` field in this request?
It was suggested in the review for https://github.com/maidsafe/safe-nd/pull/43.
So the `source` will be derived from the requester's public key.

Alternatively, don't make these changes in this PR - these can be addressed separately.

Yoga07: These changes are already made in the same PR #43 . We will rebase this once #43 gets merged.

src/append_only_data.rs
```diff
@@ -688,6 +680,167 @@ where
...
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash)]
+pub enum AData {
+    PubSeq(PubSeqAppendOnlyData),
+    UnpubSeq(UnpubSeqAppendOnlyData),
+    PubUnseq(PubUnseqAppendOnlyData),
+    UnpubUnseq(UnpubUnseqAppendOnlyData),
+}
+
+impl AData {
```
mrcnski, line 720: Can this be renamed to `Kind` and exported as `ADataKind`, like how we do with `IDataKind`?

src/append_only_data.rs
```diff
@@ -688,6 +680,167 @@ where
     }
 }
 
+macro_rules! check_perm {
+    ($data: ident, $requester: ident, $request: ident) => {
+        if $data
+            .inner
+            .owners
```
mrcnski: It would be more elegant here and future-proof to use the provided AData interface instead of accessing the internals directly. Something like this:

```rust
if $data.fetch_owner_at_index(data.owners_index()-1)...
```

Maybe you could even add a `last_owner` method and use that instead, and rename the `last` method to `last_data`.

Same for `inner.permissions.last()` below :)

src/append_only_data.rs
Yoga07: Sure @m-cat 🙂 but this post [here](https://safenetforum.org/t/rfc-54-published-and-unpublished-datatype/28620/17?u=yogesh) and it's further discussions are making me think if I should wait.

mrcnski: Ah I just saw that too. It's fine to leave this for now until we decide on the naming scheme. For the record I think `ADataKind` is more descriptive than `AData`, but maybe it's more verbose than necessary.

src/identity/mod.rs
```diff
@@ -42,6 +41,20 @@ impl PublicId {
             PublicId::App(pub_id) => pub_id.owner_name(),
         }
     }
+
+    /// Returns the PublicId serialised and encoded in z-base-32.
+    pub fn encode_to_zbase32(&self) -> String {
+        match self {
```
Fraser999: This is incorrect - we need to just call `utils::encode(&self)` so that the variant type gets encoded too.  This would get picked up if there were a basic unit test which tested the "happy" path, i.e. that decoding an encoded `PublicId` equals the initial value.

src/identity/mod.rs
```diff
@@ -73,72 +86,59 @@ struct BlsKeypairShare {
...
-
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::ClientFullId;
     use unwrap::unwrap;
 
     #[test]
-    fn encode_client_public_id_to_base64() {
+    fn encode_client_public_id_to_base32() {
```
Fraser999: These three test names should include "zbase32" rather than just "base32".

src/append_only_data.rs
```diff
@@ -166,6 +167,16 @@ impl Address {
...
         }
     }
+
+    /// Returns the PublicId serialised and encoded in z-base-32.
+    pub fn encode_to_zbase32(&self) -> String {
+        utils::encode(&self)
+    }
+
+    /// Create from z-base-32 encoded string.
+    pub fn decode_from_zbase32<T: AsRef<str>>(&self, encoded: &T) -> Result<Self> {
```
Fraser999: We have a variety of different signatures for these nine similar helper methods; some require `?Sized` and some take `&self`.  Also, some map the error needlessly.  Every one should be identical to the following I think:
```rust
/// Create from z-base-32 encoded string.
pub fn decode_from_zbase32<T: Decodable>(encoded: T) -> Result<Self> {
    utils::decode(encoded)
}
```
where `Decodable` is [`multibase::Decodable`](https://docs.rs/multibase/0.6.0/multibase/trait.Decodable.html).
(Some probably aren't using the crate's `Result` and will be returning `Result<Self, Error>`, but identical apart from that)

src/utils.rs
```diff
@@ -0,0 +1,49 @@
...
+    let message = serialise(&(request, *message_id));
+    public_key.verify(signature, message)
+}
+
+/// Wrapper for raw bincode::serialize in safe-nd
+pub(crate) fn serialise<T: Serialize>(data: &T) -> Vec<u8> {
+    unwrap!(bincode::serialize(data))
+}
+
+/// Wrapper for Base32z multibase::encode in safe-nd
```
Fraser999: "Base32z" → "z-base-32"

src/utils.rs
```diff
@@ -0,0 +1,49 @@
...
+pub(crate) fn encode<T: Serialize>(data: &T) -> String {
+    let serialised = serialise(&data);
+    multibase::encode(Base::Base32z, &serialised)
+}
+
+/// Wrapper for Base32z multibase::decode in safe-nd
```
Fraser999: "Base32z" → "z-base-32"

src/utils.rs
```diff
@@ -0,0 +1,49 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use unwrap::unwrap;
+
+/// Verify that a signature is valid for a given Request + MessageId combination
+pub fn verify_signature(
+    signature: &Signature,
+    public_key: &PublicKey,
+    request: &Request,
+    message_id: &MessageId,
+) -> Result<(), Error> {
```
Fraser999: We should import and use `crate::Result` where possible, so this becomes just `Result<()>`

src/utils.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/// Wrapper for Base32z multibase::encode in safe-nd
+pub(crate) fn encode<T: Serialize>(data: &T) -> String {
+    let serialised = serialise(&data);
+    multibase::encode(Base::Base32z, &serialised)
+}
+
+/// Wrapper for Base32z multibase::decode in safe-nd
+pub(crate) fn decode<I: ?Sized + AsRef<str>, O: DeserializeOwned>(encoded: &I) -> Result<O, Error> {
```
Fraser999: This signature should be:
```rust
pub(crate) fn decode<I: Decodable, O: DeserializeOwned>(encoded: I) -> Result<O> {
```

src/utils.rs
```diff
@@ -0,0 +1,49 @@
...
+    signature: &Signature,
+    public_key: &PublicKey,
+    request: &Request,
+    message_id: &MessageId,
+) -> Result<(), Error> {
+    let message = serialise(&(request, *message_id));
+    public_key.verify(signature, message)
+}
+
+/// Wrapper for raw bincode::serialize in safe-nd
```
Fraser999: I'm not sure "in safe-nd" adds much to the three comments?

src/public_key.rs
```diff
@@ -181,3 +185,19 @@ impl Debug for Signature {
...
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use threshold_crypto::SecretKey;
+
+    #[test]
+    fn publickey_encode_decode() {
```
Fraser999: "publickey" → "public_key"

src/lib.rs
```diff
@@ -104,6 +106,7 @@ use rand::{
 };
 use serde::{Deserialize, Serialize};
 use std::fmt::{self, Debug, Display, Formatter};
+pub use utils::verify_signature;
```
Fraser999: We have a block of public `use` statements above here - probably worth moving this up there.

src/append_only_data.rs
```diff
@@ -166,6 +168,16 @@ impl Address {
             | Address::UnpubUnseq { tag, .. } => *tag,
         }
     }
+
+    /// Returns the PublicId serialised and encoded in z-base-32.
```
Fraser999: "PublicId" → "Address"

src/immutable_data.rs
```diff
@@ -190,6 +191,16 @@ impl Address {
             Address::Pub(_) => true,
         }
     }
+
+    /// Returns the PublicId serialised and encoded in z-base-32.
```
Fraser999: "PublicId" → "Address"

src/mutable_data.rs
```diff
@@ -660,6 +661,16 @@ impl Address {
             Address::Unseq { tag, .. } | Address::Seq { tag, .. } => *tag,
         }
     }
+
+    /// Returns the PublicId serialised and encoded in z-base-32.
```
Fraser999: "PublicId" → "Address"

PR #46, opened 2019-06-28
fix/mdata: implement constructor and default trait for Mutable Data Seq/Unseq Entry actions

Yoga07:
Changes for maidsafe/safe_client_libs#841 in SCL

### DIFF
--- a/src/mutable_data.rs
+++ b/src/mutable_data.rs
@@ -393,10 +393,10 @@ impl UnseqMutableData {
 
     pub fn mutate_entries(
         &mut self,
-        actions: BTreeMap<Vec<u8>, UnseqEntryAction>,
+        actions: UnseqEntryActions,
         requester: PublicKey,
     ) -> Result<()> {
-        let (insert, update, delete) = actions.into_iter().fold(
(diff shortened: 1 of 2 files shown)


src/mutable_data.rs
```diff
@@ -689,6 +685,12 @@ pub struct SeqEntryActions {
 }
 
 impl SeqEntryActions {
+    /// Create a new Unsequenced Entry Actions list
```
mrcnski: Should say "Sequenced" here :)

src/mutable_data.rs
```diff
@@ -689,6 +685,12 @@ pub struct SeqEntryActions {
...
+    pub fn new() -> Self {
+        let actions: BTreeMap<Vec<u8>, SeqEntryAction> = Default::default();
```
mrcnski: Alternate implementation: just derive `Default` for `SeqEntryActions` (since `Btreemap` implements`Default`) and then this function can just be:

```rust
    pub fn new() -> Self {
        Default::default()
    }
```

This would save a few lines :) Same for `UnseqEntryActions` below.

PR #47, opened 2019-06-28
refactor/data: refactor data types

m-cat:
Addresses this [forum feedback](https://safenetforum.org/t/rfc-54-published-and-unpublished-datatype/28620/17), and makes the data types easier to work with in SCL (based on inefficiencies I noticed while working with the new MData types in https://github.com/maidsafe/safe_client_libs/issues/814).

### DIFF
--- a/src/mutable_data.rs
+++ b/src/mutable_data.rs
@@ -12,6 +12,8 @@ use multibase::Decodable;
 use serde::{Deserialize, Serialize};
 use std::{
     collections::{btree_map::Entry, BTreeMap, BTreeSet},
+    fmt::{self, Debug, Formatter},
+    hash::Hash,
     mem,
 };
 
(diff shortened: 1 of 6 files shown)


src/mutable_data.rs
```diff
@@ -47,8 +55,8 @@ impl Value {
     }
 }
 
-impl std::fmt::Debug for Value {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+impl Debug for Value {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
```
Fraser999, line 59: Hmm - I just noticed `std::str::from_utf8(&self.data).unwrap()` below (not part of this PR, I know).

Apart from the fact that we should always prefer the `unwrap!` macro, couldn't that cause the network to crash?  There's no requirement for the values to be valid UTF8 strings afaik, so if vaults e.g. try and log a message containing one of these, they could panic.

I think for all large binary data values we should use Andreas' [`hex_fmt` crate](https://crates.io/crates/hex_fmt), as we do [here](https://github.com/Fraser999/safe-nd/blob/master/src/lib.rs#L147) for example, or if there no real benefit in seeing the value, then just printing `..` is a good option too.

PR #49, opened 2019-07-02
Change PutAccount RPC

nbaksalyar:
Instead of a single PutAccount RPC, we now have `CreateAccount`, `CreateAccountFor`, and `UpdateAccount`. `CreateAccountFor` is required for the case when a user asks a third party to create a new coin balance _and_ a new account for them.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -7,6 +7,9 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
+mod account_data;
+
+pub use self::account_data::{AccountData, MAX_ACCOUNT_DATA_BYTES};
 use crate::{
     AData, ADataAddress, ADataAppend, ADataIndex, ADataOwner, ADataPubPermissions,
     ADataUnpubPermissions, ADataUser, AppPermissions, Coins, IDataAddress, IDataKind, MDataAddress,
@@ -15,6 +18,8 @@ use crate::{
 };
 use serde::{Deserialize, Serialize};
 use std::fmt;
+
+pub type TransactionId = u64; // TODO: Use the trait UUID
 
 /// RPC Request that is sent to vaults
 #[allow(clippy::large_enum_variant, missing_docs)]
@@ -152,12 +157,12 @@ pub enum Request {
     TransferCoins {
         destination: XorName,
         amount: Coins,
-        transaction_id: u64, // TODO: Use the trait UUID
+        transaction_id: TransactionId,
     },
     /// Get transaction
     GetTransaction {
         coins_balance_id: XorName,
-        transaction_id: u64, // TODO: Use the trait UUID
+        transaction_id: TransactionId,
     },
     /// Get current wallet balance
     GetBalance,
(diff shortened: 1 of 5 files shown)


src/request.rs
```diff
@@ -11,11 +11,21 @@ use crate::{
...
+    Signature, UnseqMutableData, XorName,
 };
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
+pub type TransactionId = u64; // TODO: Use the trait UUID
+
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+pub struct NewAccount {
+    pub destination: XorName,
```
Fraser999: I think I'd make the fields all private with public getters, since there shouldn't ever be a need to mutate any that I can think of.

src/request.rs
```diff
@@ -11,11 +11,21 @@ use crate::{
...
-    UnseqMutableData, XorName,
+    Signature, UnseqMutableData, XorName,
 };
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
+pub type TransactionId = u64; // TODO: Use the trait UUID
+
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+pub struct NewAccount {
```
Fraser999: I think we need this same struct for the `UpdateAccount` variant, so it probably could do with a different name.

src/request.rs
```diff
@@ -11,11 +11,21 @@ use crate::{
...
+    pub destination: XorName,
+    pub owner: PublicKey, // deterministically created from passwords
```
Fraser999: I wonder if we could call this something like `authorised_getter` instead to make the intent clearer?  Because this is the transient key used just for logging in really, I don't see it as representing the owner as much as his other identities with coin balances attached, etc.  I'm not sure though - it's certainly not a blocker.

src/response.rs
```diff
@@ -102,7 +102,11 @@ pub enum Response {
...
     //
     /// Returns a success or failure status of putting a new account.
-    PutAccount(Result<()>),
+    CreateAccount(Result<()>),
+    /// Returns a success or failure status of creating an account for someone else.
+    CreateAccountFor(Result<()>),
+    /// Returns a success or failure status of updating an existing account.
+    UpdateAccount(Result<()>),
     /// Returns an encrypted account packet
     GetAccount(Result<Vec<u8>>),
```
Fraser999: This could be a pair of `(Vec<u8>, Signature)`.  The client will then check the signature when the response arrives.

src/request.rs
```diff
@@ -152,25 +162,32 @@ pub enum Request {
...
         new_balance_owner: PublicKey,
         amount: Coins,
-        transaction_id: u64, // TODO: Use the trait UUID
+        transaction_id: TransactionId,
     },
     //
     // ===== Account =====
     //
-    PutAccount {
+    CreateAccount(NewAccount),
```
mrcnski: Why have a new type (`NewAccount`) just for this `CreateAccount` variant, but not for `CreateAccountFor` (same amount of fields)?

nbaksalyar: > new type (NewAccount) just for this CreateAccount variant

It's actually included as part of `CreateAccountFor` (and now `UpdateAccount` too) :slightly_smiling_face:

src/request.rs
```diff
@@ -11,11 +11,39 @@ use crate::{
...
+
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+pub struct AccountData {
+    destination: XorName,
+    authorised_getter: PublicKey, // deterministically created from passwords
+    data: Vec<u8>,                // up to 1 MB
+    signature: Signature,
+}
+
+impl AccountData {
```
Fraser999: I probably should have mentioned in the last review, but with the fields being private, we need a constructor now I think.

It might also be worth adding a pub const `MAX_ACCOUNT_DATA_BYTES` and having something like `pub fn size_is_valid(&self) -> bool` where we do `self.data.len() <= MAX_ACCOUNT_DATA_BYTES`.

This could be called inside the constructor and return an error if it's false.

We could even add that check when deserialising and likewise return an error, then `size_is_valid` wouldn't even need to be `pub`.  We'd be guaranteed that every instance of `AccountData` is validly-sized.  However, implementing `Deserialize` by hand is maybe not worth the risk.

PR #50, opened 2019-07-02
fix/adata: fix pushing duplicate entries in AppendOnly Data

Yoga07:
Resolves maidsafe/safe_client_libs#875

### DIFF
--- a/src/append_only_data.rs
+++ b/src/append_only_data.rs
@@ -11,7 +11,7 @@ use crate::{utils, Error, PublicKey, Request, Result, XorName};
 use multibase::Decodable;
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::BTreeMap,
+    collections::{BTreeMap, BTreeSet},
     fmt::{self, Debug, Formatter},
     hash::Hash,
 };
@@ -343,7 +343,7 @@ pub trait AppendOnlyData<P> {
 /// Common methods for published and unpublished unsequenced `AppendOnlyData`.
 pub trait UnseqAppendOnly {
     /// Append new entries.
-    fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)]) -> Result<()>;
+    fn append(&mut self, entries: Entries) -> Result<()>;
 }
 
 /// Common methods for published and unpublished sequenced `AppendOnlyData`.
@@ -352,7 +352,7 @@ pub trait SeqAppendOnly {
     ///
     /// If the specified `last_entries_index` does not match the last recorded entries index, an
     /// error will be returned.
-    fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)], last_entries_index: u64) -> Result<()>;
+    fn append(&mut self, entries: Entries, last_entries_index: u64) -> Result<()>;
 }
 
 macro_rules! impl_appendable_data {
@@ -632,15 +632,37 @@ impl Debug for UnseqAppendOnlyData<UnpubPermissions> {
     }
 }
 
+fn check_dup(data: &[(Vec<u8>, Vec<u8>)], entries: &mut Entries) -> Result<()> {
+    let new: BTreeSet<&Vec<u8>> = entries.iter().map(|(key, _value)| key).collect();
+
+    // If duplicate entries are present in the push.
+    if new.len() < entries.len() {
+        return Err(Error::DuplicateEntryKeys);
+    }
+
+    let existing: BTreeSet<&Vec<u8>> = data.iter().map(|(key, _value)| key).collect();
+    if !existing.is_disjoint(&new) {
+        let dup: Entries = entries
+            .drain(..)
+            .filter(|(key, _value)| existing.contains(&key))
+            .collect();
+        return Err(Error::KeysExist(dup));
+    }
+    Ok(())
+}
+
 impl<P> SeqAppendOnly for SeqAppendOnlyData<P>
 where
     P: Permissions + Hash + Clone,
 {
-    fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)], last_entries_index: u64) -> Result<()> {
+    fn append(&mut self, mut entries: Entries, last_entries_index: u64) -> Result<()> {
+        check_dup(&self.inner.data, entries.as_mut())?;
+
         if last_entries_index != self.inner.data.len() as u64 {
             return Err(Error::InvalidSuccessor(self.inner.data.len() as u64));
         }
-        self.inner.data.extend(entries.iter().cloned());
+
+        self.inner.data.extend(entries);
         Ok(())
     }
 }
@@ -649,8 +671,10 @@ impl<P> UnseqAppendOnly for UnseqAppendOnlyData<P>
 where
     P: Permissions + Hash + Clone,
 {
-    fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)]) -> Result<()> {
-        self.inner.data.extend(entries.iter().cloned());
+    fn append(&mut self, mut entries: Entries) -> Result<()> {
+        check_dup(&self.inner.data, entries.as_mut())?;
+
+        self.inner.data.extend(entries);
         Ok(())
     }
 }
@@ -899,7 +923,7 @@ pub struct AppendOperation {
 mod tests {
     use super::*;
     use threshold_crypto::SecretKey;
-    use unwrap::unwrap;
+    use unwrap::{unwrap, unwrap_err};
 
     #[test]
     fn append_permissions() {
@@ -988,12 +1012,7 @@ mod tests {
     #[test]
     fn seq_append_entries() {
         let mut data = SeqAppendOnlyData::<PubPermissions>::new(XorName([1; 32]), 10000);
-        let res = data.append(&[(b"hello".to_vec(), b"world".to_vec())], 0);
-
-        match res {
-            Ok(()) => (),
-            Err(x) => panic!("Unexpected error: {:?}", x),
-        }
+        unwrap!(data.append(vec![(b"hello".to_vec(), b"world".to_vec())], 0));
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


src/append_only_data.rs
```diff
@@ -676,6 +698,28 @@ where
     P: Permissions + std::hash::Hash + Clone,
 {
     fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)]) -> Result<()> {
+        let existing: BTreeSet<Vec<u8>> = self
```
lionel-faber: This block seems to be the same as the previous one.
We can consider putting it into a function.

src/errors.rs
```diff
@@ -40,6 +41,10 @@ pub enum Error {
     InvalidEntryActions(BTreeMap<Vec<u8>, EntryError>),
     /// Key does not exist
     NoSuchKey,
+    /// Key already exists
+    KeyExists(Entries),
```
lionel-faber: I think this is better named as `KeysExist`

src/errors.rs
```diff
@@ -40,6 +41,10 @@ pub enum Error {
...
+    /// Duplicate Entries in this push
+    DuplicateEntries,
```
lionel-faber: This error is returned if the _keys_ are duplicate.

src/append_only_data.rs
```diff
@@ -663,6 +663,28 @@ where
     P: Permissions + std::hash::Hash + Clone,
 {
     fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)], last_entries_index: u64) -> Result<()> {
+        let existing: BTreeSet<Vec<u8>> = self
```
Fraser999: `existing` isn't used until line 679 below, so better to create it just immediately before there rather than this early.  (e.g. this would would be wasted effort if we return at line 676).

src/append_only_data.rs
```diff
@@ -663,6 +663,28 @@ where
...
+            .inner
+            .data
+            .iter()
+            .map(|tuple| tuple.0.clone())
```
Fraser999: I think we should be able to avoid cloning the keys of the entries if we make the type of `existing` and `new` `BTreeSet<&Vec<u8>>`.

Also, a minor point: to make the code potentially more readable, you could change the captured variable from `tuple` to `(key, _value)`, then rather than calling `tuple.0` you'd just have `key`.  Not a big deal for something as simple as this scenario though.

src/append_only_data.rs
```diff
@@ -663,6 +663,28 @@ where
     P: Permissions + std::hash::Hash + Clone,
 {
     fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)], last_entries_index: u64) -> Result<()> {
```
Fraser999: I think it may be better to just pass a `Vec` here rather than a slice, since the collection will be consumed by this function.  I.e. changing the signature to `mut entries: Entries`.

src/append_only_data.rs
```diff
@@ -663,6 +663,28 @@ where
...
+            .collect();
+        let new: BTreeSet<Vec<u8>> = entries.iter().map(|tuple| tuple.0.clone()).collect();
+
+        // If duplicate entries are present in the push.
+        if new.len() < entries.len() {
+            return Err(Error::DuplicateEntries);
+        }
+
+        if !existing.is_disjoint(&new) {
+            let dup: Entries = entries
```
Fraser999: If we do take `entries` in as a `mut Vec`, we can avoid cloning here and just do:
```rust
let dup = entries
    .drain(..)
    .filter(|(key, _value)| existing.contains(&key))
    .collect();
```
and likewise at the end of the method, we could just do
```rust
self.inner.data.extend(entries);
```

src/errors.rs
Fraser999: I guess to avoid confusion with crypto keys though, this could maybe be renamed to `DuplicateEntryKeys` rather than just `DuplicateKeys`?

src/append_only_data.rs
```diff
@@ -658,15 +658,39 @@ impl UnseqAppendOnlyData<UnpubPermissions> {
...
+    Ok(())
+}
+
 impl<P> SeqAppendOnly for SeqAppendOnlyData<P>
 where
     P: Permissions + std::hash::Hash + Clone,
 {
-    fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)], last_entries_index: u64) -> Result<()> {
+    fn append(&mut self, mut entries: Entries, last_entries_index: u64) -> Result<()> {
+        match check_dup(&self.inner.data, entries.as_mut()) {
```
nbaksalyar: Probably can be replaced with just
```suggestion
        if let Err(e) = check_dup(&self.inner.data, entries.as_mut()) {
            return Err(e);
        }
```

src/append_only_data.rs
```diff
@@ -675,8 +699,13 @@ impl<P> UnseqAppendOnly for UnseqAppendOnlyData<P>
 where
     P: Permissions + std::hash::Hash + Clone,
 {
-    fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)]) -> Result<()> {
-        self.inner.data.extend(entries.iter().cloned());
+    fn append(&mut self, mut entries: Entries) -> Result<()> {
+        match check_dup(&self.inner.data, entries.as_mut()) {
```
nbaksalyar: Same here, I think `if let` should suffice?

src/append_only_data.rs
```diff
@@ -969,4 +998,15 @@ mod tests {
...
+
+    #[test]
+    fn append_unseq_test() {
+        let mut data = UnpubUnseqAppendOnlyData::new(XorName(rand::random()), 10);
+        let key1 = b"KEY1".to_vec();
+        let val1 = b"VALUE1".to_vec();
+        let key2 = b"KEY2".to_vec();
+        let val2 = b"VALUE2".to_vec();
+        let tup = [(key1, val1), (key2, val2)].to_vec();
+        assert_eq!(unwrap!(data.append(tup)), ());
```
nbaksalyar: Can be
```suggestion
        assert!(data.append(tup).is_ok());
```

src/append_only_data.rs
Fraser999: Or even just
```rust
check_dup(&self.inner.data, entries.as_mut())?;
```

src/append_only_data.rs
```diff
@@ -931,7 +960,7 @@ mod tests {
     #[test]
     fn seq_append_entries() {
         let mut data = SeqAppendOnlyData::<PubPermissions>::new(XorName([1; 32]), 10000);
-        let res = data.append(&[(b"hello".to_vec(), b"world".to_vec())], 0);
+        let res = data.append([(b"hello".to_vec(), b"world".to_vec())].to_vec(), 0);
```
Fraser999: A minor point: this could be `vec![...]` rather than `[...].to_vec()`.

Also, we seem to have a few instances of
```rust
match res {
    Ok(()) => (),
    Err(x) => panic!("Unexpected error: {:?}", x),
}
```
These could all be removed if we just do e.g.
```rust
unwrap!(data.append(vec![(b"hello".to_vec(), b"world".to_vec())], 0));
```

src/append_only_data.rs
Fraser999: I agree the `assert_eq` in unnecessary, but I generally prefer unwrapping results in tests rather than just checking `is_ok()` since if it fails, you see the error value.  Not a big deal though.

src/errors.rs
```diff
@@ -40,6 +41,10 @@ pub enum Error {
     InvalidEntryActions(BTreeMap<Vec<u8>, EntryError>),
     /// Key does not exist
     NoSuchKey,
+    /// Key already exists
+    KeysExists(Entries),
```
Fraser999: Should be `KeysExist`.  Also, the doc comment could probably be expanded a bit, e.g.
```rust
/// The key(s) of the entry or entries contained in this error already exist
```

src/append_only_data.rs
```diff
@@ -658,15 +658,38 @@ impl UnseqAppendOnlyData<UnpubPermissions> {
...
+    Ok(())
+}
+
 impl<P> SeqAppendOnly for SeqAppendOnlyData<P>
 where
     P: Permissions + std::hash::Hash + Clone,
 {
-    fn append(&mut self, entries: &[(Vec<u8>, Vec<u8>)], last_entries_index: u64) -> Result<()> {
+    fn append(&mut self, mut entries: Entries, last_entries_index: u64) -> Result<()> {
+        if let Err(e) = check_dup(&self.inner.data, entries.as_mut()) {
```
Fraser999: This can be just:
```rust
check_dup(&self.inner.data, entries.as_mut())?;
```
Applies in another place below too.

src/append_only_data.rs
```diff
@@ -969,4 +996,64 @@ mod tests {
...
+        let mut data = UnpubUnseqAppendOnlyData::new(XorName(rand::random()), 10);
+
+        // Assert that the current Append Operation has duplicate entries present and are not pushed
+        let entries = [
+            (b"KEY1".to_vec(), b"VALUE1".to_vec()),
+            (b"KEY2".to_vec(), b"VALUE2".to_vec()),
+            (b"KEY1".to_vec(), b"VALUE1".to_vec()),
+        ]
+        .to_vec();
+        assert!(data.append(entries).is_err());
```
Fraser999: It'd be more thorough to check we're getting the correct variant returned:
```rust
assert_eq!(Error::DuplicateEntryKeys, unwrap_err!(data.append(entries)));
```
Also, it might be worth checking the state of `data.entries()` is as expected?  Even just once at the very end of the test, or after each attempt to mutate the data.

src/append_only_data.rs
```diff
@@ -1026,4 +1045,70 @@ mod tests {
         let decoded = unwrap!(self::Address::decode_from_zbase32(&encoded));
         assert_eq!(address, decoded);
     }
+
+    #[test]
+    fn append_unseq_data_test() {
+        let mut data = UnpubUnseqAppendOnlyData::new(XorName(rand::random()), 10);
+
+        // Assert that the current Append Operation has duplicate entries present and are not pushed
```
nbaksalyar: Maybe rephrase it a bit? Also, we're looking for duplicate _keys_ as opposed to entries (keys + values).
```suggestion
        // Assert that the entries are not appended because of duplicate keys.
```

src/append_only_data.rs
```diff
@@ -1026,4 +1045,70 @@ mod tests {
...
+        let entries = vec![
+            (b"KEY1".to_vec(), b"VALUE1".to_vec()),
+            (b"KEY2".to_vec(), b"VALUE2".to_vec()),
+            (b"KEY1".to_vec(), b"VALUE1".to_vec()),
+        ];
+        assert_eq!(Error::DuplicateEntryKeys, unwrap_err!(data.append(entries)));
+
+        // Assert that the current Append Operation has no duplicate entries present and are pushed
```
nbaksalyar: Maybe rephrase it a bit? Also, we're looking for duplicate keys as opposed to entries (keys + values).
```suggestion
        // Assert that the entries are appended because there are no duplicate keys.
```

src/append_only_data.rs
```diff
@@ -1026,4 +1045,70 @@ mod tests {
...
+        let entries1 = vec![
+            (b"KEY1".to_vec(), b"VALUE1".to_vec()),
+            (b"KEY2".to_vec(), b"VALUE2".to_vec()),
+        ];
+
+        unwrap!(data.append(entries1));
+
+        // Assert that duplicate entries present and are not pushed
```
nbaksalyar:
```suggestion
        // Assert that entries are not appended because they duplicate some keys appended previously.
```

src/append_only_data.rs
```diff
@@ -1026,4 +1045,70 @@ mod tests {
...
+        let entries2 = vec![(b"KEY2".to_vec(), b"VALUE2".to_vec())];
+        assert_eq!(
+            Error::KeysExist(entries2.clone()),
+            unwrap_err!(data.append(entries2))
+        );
+
+        // Assert that duplicate entries not present and are pushed
```
nbaksalyar:
```suggestion
        // Assert that no duplicate keys are present and the append operation is successful.
```

src/append_only_data.rs
```diff
@@ -1026,4 +1045,70 @@ mod tests {
...
+        let entries3 = vec![(b"KEY3".to_vec(), b"VALUE3".to_vec())];
+        unwrap!(data.append(entries3));
+    }
+
+    #[test]
+    fn append_seq_data_test() {
+        let mut data = UnpubSeqAppendOnlyData::new(XorName(rand::random()), 10);
+
+        // Assert that the current Append Operation has duplicate entries present and are not pushed
```
nbaksalyar:
```suggestion
        // Assert that the entries are not appended because of duplicate keys.
```

src/append_only_data.rs
```diff
@@ -1026,4 +1045,70 @@ mod tests {
...
+            (b"KEY1".to_vec(), b"VALUE1".to_vec()),
+            (b"KEY2".to_vec(), b"VALUE2".to_vec()),
+            (b"KEY1".to_vec(), b"VALUE1".to_vec()),
+        ];
+        assert_eq!(
+            Error::DuplicateEntryKeys,
+            unwrap_err!(data.append(entries, 0))
+        );
+
+        // Assert that the current Append Operation has no duplicate entries present and are pushed
```
nbaksalyar:
```suggestion
        // Assert that the entries are appended because there are no duplicate keys.
```

src/append_only_data.rs
```diff
@@ -1026,4 +1045,70 @@ mod tests {
...
+        let entries1 = vec![
+            (b"KEY1".to_vec(), b"VALUE1".to_vec()),
+            (b"KEY2".to_vec(), b"VALUE2".to_vec()),
+        ];
+        unwrap!(data.append(entries1, 0));
+
+        // Assert that duplicate entries present and are not pushed
```
nbaksalyar:
```suggestion
        // Assert that entries are not appended because they duplicate some keys appended previously.
```

src/append_only_data.rs
```diff
@@ -1026,4 +1045,70 @@ mod tests {
...
+        let entries2 = vec![(b"KEY2".to_vec(), b"VALUE2".to_vec())];
+        assert_eq!(
+            Error::KeysExist(entries2.clone()),
+            unwrap_err!(data.append(entries2, 2))
+        );
+
+        // Assert that duplicate entries not present and are pushed
```
nbaksalyar:
```suggestion
        // Assert that no duplicate keys are present and the append operation is successful.
```

PR #54, opened 2019-07-04
fix/idata: make unpub idata owner field use safe-nd PublicKey

Yoga07:
Hey guys, 

I am not sure of the `.to_bytes()` impl for `PublicKey` here. The default `to_bytes()` of ed25519 key gives a `[u8; 32]` whereas, thresholdcrypto keys' `to_bytes()` gives a `[u8,48]` . Hence I had to convert it to Vector for returning and again make it a slice at the consumption point. Can this be implemented a better way?(I am sure it can be :) )

Fixes https://github.com/maidsafe/safe_client_libs/issues/882

### DIFF
--- a/src/immutable_data.rs
+++ b/src/immutable_data.rs
@@ -7,7 +7,7 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
-use crate::{utils, Error, XorName, XOR_NAME_LEN};
+use crate::{utils, Error, PublicKey, XorName, XOR_NAME_LEN};
 use bincode::serialized_size;
 use multibase::Decodable;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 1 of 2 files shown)


src/immutable_data.rs
```diff
@@ -38,7 +38,7 @@ impl UnpubImmutableData {
         // TODO: Use low-level arrays or slices instead of Vec.
         let mut bytes = Vec::with_capacity(XOR_NAME_LEN + PK_SIZE);
         bytes.extend_from_slice(&tiny_keccak::sha3_256(&value));
-        bytes.extend_from_slice(&owners.to_bytes());
+        bytes.extend_from_slice(&owners.to_bytes()[..48]);
```
mrcnski, line 41: Where does `48` come from? Can we use a descriptive constant here instead of hard-coding a number?

Fraser999: I think rather than writing a new `to_bytes()` method we should just use the existing ability to serialise the `PublicKey`.  Deriving the address could be something like:
```rust
let hash_of_value = tiny_keccak::sha3_256(&value);
let serialised_contents = utils::serialise(&(hash_of_value, &owners));
let address = Address::Unpub(XorName(tiny_keccak::sha3_256(&serialised_contents)));
```

mrcnski:
Closing in favor of https://github.com/maidsafe/safe-nd/pull/58.

PR #58, opened 2019-07-08
fix/idata: make unpub idata owner field use safe-nd PublicKey

m-cat:
Fixes maidsafe/safe_client_libs#882

### DIFF
--- a/src/immutable_data.rs
+++ b/src/immutable_data.rs
@@ -7,7 +7,7 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
-use crate::{utils, Error, XorName, XOR_NAME_LEN};
+use crate::{utils, Error, PublicKey, XorName};
 use bincode::serialized_size;
 use multibase::Decodable;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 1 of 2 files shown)


src/immutable_data.rs
```diff
@@ -35,11 +34,10 @@ pub struct UnpubImmutableData {
 impl UnpubImmutableData {
     /// Creates a new instance of `UnpubImmutableData`
     pub fn new(value: Vec<u8>, owners: PublicKey) -> Self {
```
nbaksalyar: While we are at it, could we please change `owners` to just `owner`? Since we are agnostic to the key type now.
```suggestion
    pub fn new(value: Vec<u8>, owner: PublicKey) -> Self {
```
Same for the field name in struct and the getter function.

src/public_key.rs
```diff
@@ -51,6 +51,14 @@ impl PublicKey {
     pub fn decode_from_zbase32<I: Decodable>(encoded: I) -> Result<Self> {
         utils::decode(encoded)
     }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
```
Fraser999: Do we still need this method?

It seems a little risky since we erase the information about which variant has been serialised.  Also, the `PublicKey` type already derives `Serialize`, so I don't see a use case for this `to_bytes()` method.

PR #60, opened 2019-07-09
Rename account-related structs to use "LoginPacket"

lionel1704:


src/request.rs
```diff
@@ -172,17 +172,17 @@ pub enum Request {
...
     //
-    // ===== Account =====
+    // ===== Login Packet =====
     //
-    CreateAccount(AccountData),
-    CreateAccountFor {
-        new_account_owner: PublicKey,
+    CreateLoginPacket(LoginPacket),
+    CreateLoginPacketFor {
+        new_packet_owner: PublicKey,
```
nbaksalyar: Maybe just `new_owner`?

src/request/login_packet.rs
```diff
@@ -10,40 +10,40 @@
 use crate::{Error, PublicKey, Result, Signature, XorName};
 use serde::{Deserialize, Serialize};
 
-/// Account packet size is limited .
-pub const MAX_ACCOUNT_DATA_BYTES: usize = 1024 * 1024; // 1 MB
+/// Login packet size is limited .
```
nbaksalyar, line 13: I think we should change this comment to something like
```suggestion
/// Login packet size has a limit of 1 MB (not including the address, signature, and a key).
```
But can be addressed in a separate PR to add missing documentation.

PR #63, opened 2019-07-10
feat/adata: add GetADataValue request/response types and get_value function for AData enum

Yoga07:
Required for maidsafe/safe_client_libs#879

### DIFF
--- a/src/append_only_data.rs
+++ b/src/append_only_data.rs
@@ -829,6 +829,15 @@ impl Data {
         }
     }
 
+    pub fn get(&self, key: &[u8]) -> Option<&Vec<u8>> {
+        match self {
+            Data::PubSeq(data) => data.get(key),
+            Data::PubUnseq(data) => data.get(key),
+            Data::UnpubSeq(data) => data.get(key),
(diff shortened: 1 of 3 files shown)


src/append_only_data.rs
```diff
@@ -805,6 +805,15 @@ impl Data {
         }
     }
 
+    pub fn get_value(&self, key: Vec<u8>) -> Option<&Vec<u8>>{
+        match self {
+            Data::PubSeq(data) => data.get(key.as_slice()),
```
nbaksalyar: I'd suggest renaming `data.get` to `data.get_value` for consistency.

src/append_only_data.rs
```diff
@@ -829,6 +829,15 @@ impl Data {
         }
     }
 
+    pub fn get(&self, key: Vec<u8>) -> Option<&Vec<u8>> {
```
nbaksalyar: This should be
```suggestion
    pub fn get(&self, key: &[u8]) -> Option<&Vec<u8>> {
```
and we wouldn't need to call `.as_slice()` for `key` then

PR #69, opened 2019-07-11
fix/value: remove unsafe `unwrap` and reimplement Debug for Value

### LINKED ISSUE #52: Remove `unwrap` and possible network crash
From @Fraser999 's review comment [here](https://github.com/maidsafe/safe-nd/pull/47#pullrequestreview-257922105):

> Hmm - I just noticed `std::str::from_utf8(&self.data).unwrap()` below (not part of this PR, I know).
> 
> Apart from the fact that we should always prefer the `unwrap!` macro, couldn't that cause the network to crash? There's no requirement for the values to be valid UTF8 strings afaik, so if vaults e.g. try and log a message containing one of these, they could panic.
> 
> I think for all large binary data values we should use Andreas' [`hex_fmt` crate](https://crates.io/crates/hex_fmt), as we do [here](https://github.com/Fraser999/safe-nd/blob/master/src/lib.rs#L147) for example, or if there no real benefit in seeing the value, then just printing `..` is a good option too.

m-cat:
Fixes #52

### DIFF
--- a/src/mutable_data.rs
+++ b/src/mutable_data.rs
@@ -8,6 +8,7 @@
 // Software.
 
 use crate::{utils, EntryError, Error, PublicKey, Request, Result, XorName};
+use hex_fmt::HexFmt;
 use multibase::Decodable;
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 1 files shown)


nbaksalyar:
Nice one, thanks!

PR #75, opened 2019-07-18
Fix crash in check_permissions, add check_is_owner

octol:


src/append_only_data.rs
```diff
@@ -467,6 +470,21 @@ macro_rules! impl_appendable_data {
                 self.inner.owners.push(owner);
                 Ok(())
             }
+
+            fn check_is_last_owner(&self, requester: PublicKey) -> Result<()> {
+                if self.owners_index() == 0 {
```
mrcnski: Minor optimization: You can store the result of `self.owners_index()` in a variable to avoid calling the function twice. Same in `check_perm` below.

madadam: You can also do just `self.owner(Index::FromEnd(1))` and avoid the `if` / `else` altogether. Same in `check_perm!` below.

octol: That would give a different error though, but maybe that's fine?

madadam: It gives you `None` which you can map to any error you want with `.or_or(...)`.

octol: You changing the existing `.ok_or_else` to `InvalidOwners` when `owners_index == 0`, do you mean having another `ok_or`? How do you separate the two cases without an `if`?

madadam: Ah I see. There are two different errors: `InvalidOwners` and `NoSuchData`. I wonder why that is? Can we return `InvalidOwners` in both cases? It seems weird to return `NoSuchData`. If we could do that, when you could do just:

```rust
let last_owner = self.owner(Index::FromEnd(1)).ok_or(Error::InvalidOwners)?;
if last_owner.public_key == requester {
    Ok(())
} else {
    Err(Error::AccessDenied)
}
```

octol: Yep if having `InvalidOwners` for both is fine then that is a clean solution! Similarly in the second branch in `check_perm` that you suggested fixing, there currently is no `InvalidPermissions` so maybe keep `NoSuchData` there?

madadam: Sure. Basically just use the same error in both cases because I see no reason to distinguish them - both errors are caused by there being no entry at the given index.

madadam: That said, maybe we should add `InvalidPermissions` error as well?

octol: I see no reason not to

PR #76, opened 2019-07-19
feat/response: add TryFrom impls for Response types

Fraser999:


src/response.rs
```diff
@@ -72,7 +76,58 @@ pub enum Response {
...
+try_from!(Vec<Vec<u8>>, ListUnseqMDataValues);
+try_from!(MDataPermissionSet, ListMDataUserPermissions);
+try_from!(BTreeMap<PublicKey, MDataPermissionSet>, ListMDataPermissions);
+try_from!(MDataValue, GetSeqMDataValue);
+try_from!(Vec<u8>, GetUnseqMDataValue, GetADataValue);
+try_from!(AData, GetAData, GetADataShell);
+try_from!(ADataOwner, GetADataOwners);
+try_from!(ADataEntries, GetADataRange);
+try_from!(ADataIndices, GetADataIndices);
+try_from!((Vec<u8>, Vec<u8>), GetADataLastEntry);
```
mrcnski, line 118: Would be nice to define a type `Entry = (Vec<u8>, Vec<u8>)` in append_only_data.rs, exporting it as `ADataEntry`.

Fraser999: If we're exporting that type anyway, do you mind if I just make it a proper struct, like
```rusr
struct Entry {
    pub key: Vec<u8>,
    pub value: Vec<u8>,
}
```

mrcnski: Sure, that's even better than my suggestion 👍

src/append_only_data.rs
```diff
@@ -398,17 +410,17 @@ macro_rules! impl_appendable_data {
...
+                    }
+                })
             }
 
-            fn last_entry(&self) -> Option<(Vec<u8>, Vec<u8>)> {
-                match self.inner.data.last() {
-                    Some(tup) => Some(tup.clone()),
-                    None => None,
-                }
+            fn last_entry(&self) -> Option<Entry> {
```
madadam: This is outside of the scope of this PR, but I wonder if we could change this to return reference instead of clone, to be consistent with `get`? The caller can always clone it themselves if needed.

Fraser999: I'd be happy to make that change if others agree.

src/append_only_data.rs
```diff
@@ -242,6 +242,18 @@ pub struct Owner {
...
 }
 
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, Default, Debug)]
+pub struct Entry {
+    pub key: Vec<u8>,
+    pub value: Vec<u8>,
+}
+
+impl Entry {
+    pub fn new(key: Vec<u8>, value: Vec<u8>) -> Self {
```
mrcnski, line 252: Looks good! I just don't understand why we want `Entry::new`, the fields are public so you can just do `Entry { key: X, value: Y }` which is a bit more explicit and not much more verbose.

src/append_only_data.rs
mrcnski: Agreed with @madadam

src/append_only_data.rs
Fraser999: Just makes it slightly more ergonomic to use.  For example, see all the changes in the tests which would have been a bit more awkward without the constructor.  E.g.
```rust
let entries = vec![
    Entry {
        key: b"KEY1".to_vec(),
        value: b"VALUE1".to_vec(),
    },
    Entry {
        key: b"KEY2".to_vec(),
        value: b"VALUE2".to_vec(),
    },
    Entry {
        key: b"KEY1".to_vec(),
        value: b"VALUE1".to_vec(),
    },
];
```
vs
```rust
let entries = vec![
    Entry::new(b"KEY1".to_vec(), b"VALUE1".to_vec()),
    Entry::new(b"KEY2".to_vec(), b"VALUE2".to_vec()),
    Entry::new(b"KEY1".to_vec(), b"VALUE1".to_vec()),
];
```

PR #80, opened 2019-07-23
fix/naming: fix up some names for consistency

m-cat:


m-cat:
The latest commit adds a new error in case of a type mismatch. This mismatch was impossible before this PR (we had two request types for seq and unseq separately, not one request for both), so I'm wondering if a slightly simplified API is worth the increased potential for errors. I'm leaning towards closing this PR.

src/mutable_data.rs
```diff
@@ -714,7 +714,7 @@ impl Data {
         }
     }
 
-    pub fn owner(&self) -> PublicKey {
+    pub fn owners(&self) -> PublicKey {
```
nbaksalyar: This should stay as `owner` because generally we shouldn't care if it's a multisig/BLS key or some other key.
It'd be better to change the `.owners` field name in `SeqMutableData` and `UnseqMutableData`.

src/mutable_data.rs
```diff
@@ -29,9 +29,9 @@ pub struct SeqMutableData {
     permissions: BTreeMap<PublicKey, PermissionSet>,
     /// Version should be increased for any changes to MutableData fields except for data.
     version: u64,
-    /// Contains a set of owners of this data. DataManagers enforce that a mutation request is
-    /// coming from the MaidManager Authority of the Owner.
-    owners: PublicKey,
+    /// Contains the owner of this data. DataManagers enforce that a mutation request is coming from
```
nbaksalyar: Just noticed that this comment mentions legacy names from Vaults.
Could you please change it to something like
```rust
/// Contains the public key of an owner or owners of this data.
/// Data Handlers in vaults enforce that a mutation request has a valid signature of the owner.
```
?

src/mutable_data.rs
```diff
@@ -67,9 +67,9 @@ pub struct UnseqMutableData {
...
```
nbaksalyar: Same comment here.

nbaksalyar:
Before we merge this, @m-cat could you please raise a PR for SAFE Client Libs to make it compatible with these changes? There are some places in Mock Vault where we use the `.owners()` name. SAFE Vault is confirmed to be working correctly with this PR integrated.

mrcnski:
Sure, I am just waiting for the tiny-keccak compatibility fixes as I can't compile and test SCL right now, it seems.

PR #82, opened 2019-07-23
Pass PublicKey by reference

madadam:


mrcnski:
Why is this change necessary if `PublicKey` is a `Copy` type?

madadam:
> Why is this change necessary if PublicKey is a Copy type?

I think passing by reference is more efficient in this case, as `PublicKey` is relatively large type (152 bytes on my machine). Might not be that significant in practice though.

mrcnski:
Makes sense, but in that case maybe it shouldn't derive `Copy`.

madadam:
> Makes sense, but in that case maybe it shouldn't derive Copy.

I see no harm in it remaining `Copy`, but I don't have strong opinion about it, so if others agree I'm fine with making the change.

Fraser999:
I'd prefer to leave it `Copy` since the [Rust docs](https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy) state:
> Generally speaking, if your type _can_ implement `Copy`, it should.

I've been meaning to benchmark this for a while now, since I'd generally expect passing a large-ish object by ref would be better than copying it.  But I don't actually know if this is true or not.

mrcnski (dismissed):
Thanks for the clarifications 👍

Fraser999 (dismissed):

lionel-faber:
@oetyng could you please pick up these changes in the new `Sequence` and `Map` types?
I'll leave this PR open for now and close it later once we employ this approach in the new types.

PR #83, opened 2019-07-23
refactor/response: add Request::error_response and simplify Response variants

m-cat:


src/request.rs
```diff
@@ -206,9 +206,74 @@ pub enum Request {
...
+            // Login Packet
+            CreateLoginPacket { .. } => Response::Mutation(Err(error)),
+            CreateLoginPacketFor { .. } => Response::Mutation(Err(error)),
+            UpdateLoginPacket { .. } => Response::Mutation(Err(error)),
+            GetLoginPacket(..) => Response::GetLoginPacket(Err(error)),
+            // Client (Owner) to SrcElders
+            ListAuthKeysAndVersion => Response::ListAuthKeysAndVersion(Err(error)),
+            InsAuthKey { .. } => Response::Mutation(Err(error)),
+            DelAuthKey { .. } => Response::Mutation(Err(error)),
+        }
```
madadam, line 273: You can reduce the amount of code here a bit by grouping all the requests that map to `Response::Mutation`:

```rust
Request::PutIData(_)
    | Request::DeleteUnpubIData(_)
    ...
    | Request::DelAuthKey { .. } => Response::Mutation(Err(error)),
```

PR #85, opened 2019-07-29
Remove limit check from Coins::from_nano

madadam:
- Remove the max-value check in `Coins::from_nano` (it now returns `Self` instead of `Result<Self>`)
- Make `Coins::from_nano` a `const fn` to simplify creating `Coins` `const`s (lazy-static not reqiured anymore). 
- Add `Coins::validate` to perform the max-value check explicitly.


src/coins.rs
```diff
@@ -30,35 +30,43 @@ pub struct Coins(u64);
...
     // Computes self - rhs, returning None if overflow occurred.
     pub fn checked_sub(self, rhs: Coins) -> Option<Coins> {
         self.0
             .checked_sub(rhs.0)
-            .and_then(|nano| Coins::from_nano(nano).ok())
+            .map(Self::from_nano)
+            .and_then(|coins| coins.validate().ok())
+    }
+
+    /// Check that the coins amount is valid - thas is - not exceeding max allowed value.
```
Fraser999: "thas" → "that"

Fraser999 (commented):
I think `validate()` and `MAX_COINS_VALUE` should be removed altogether, since it seems a bit strange to me that we can do
```rust
assert!(Coins::from_nano(4_294_967_296_000_000_000)
    .checked_sub(Coins::from_nano(0))
    .is_none());
```

Fraser999 (dismissed):

lionel-faber (approved):

PR #86, opened 2019-07-29
derive Ord, PartialOrd and Hash for PublicKey and Signature

madadam:
Instead of implementing them. This is possible because ed25519-dalek::{PublicKey,Signature} implement those traits now.

Note: this uses [git dependency](https://github.com/madadam/ed25519-dalek.git) of ed25519-dalek for now until https://github.com/dalek-cryptography/ed25519-dalek/pull/93 is merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,8 @@ version = "0.1.0"
 [dependencies]
 # Ensure bincode version is identical to that in SAFE Client Libs and SAFE Vault.
 bincode = "=1.1.4"
-ed25519-dalek = "~0.9.1"
+# ed25519-dalek = "~0.9.1"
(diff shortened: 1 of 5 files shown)


nbaksalyar:
> this uses git dependency of ed25519-dalek

I would prefer to delay merging this PR until after we publish a first version of safe-nd on crates.io.

Fraser999 (dismissed):

lionel-faber:
I see that the linked PR has been closed. So we might need to implement this in safe-nd after all?
Do you mind if I close this PR @madadam ?

madadam:
> I see that the linked PR has been closed. So we might need to implement this in safe-nd after all?
> Do you mind if I close this PR @madadam ?

Sure, makes sense. Go ahead.

PR #95, opened 2019-08-05
doc/all: add missing documentation

m-cat:
Fixes https://github.com/maidsafe/safe_client_libs/issues/954

### DIFF
--- a/src/append_only_data.rs
+++ b/src/append_only_data.rs
@@ -7,6 +7,56 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
+//! AppendOnlyData
+//!
+//! AppendOnlyData can be either published or unpublished and either sequenced or unsequenced.
+//!
+//! ## Published data
+//!
+//! Published data refers to the content that is published (made available) for everyone. For
+//! example: websites, blogs, or research papers which anyone can fetch from the network and read
+//! without requiring any permission. For such public content, it becomes important to retain a
+//! history of changes. History is not be allowed to be tampered with and the published data remains
+//! forever.
+//!
+//! The AppendOnly data type is pivotal for data perpetuity, because it ensures the published
+//! versions are always available unlike the default behaviour of MutableData where the value can be
+//! overwritten. This is central to prevent censorship of information.
+//!
+//! Data perpetuity is one of the fundamentals of the SAFE Network, which ensures the users of the
+//! network shall be able to store published data in perpetuity.
+//!
+//! However, only the owners or permitted users shall be able to append the changes based on the
+//! permission.
+//!
+//! ## Unpublished data
+//!
+//! Not all the data is desired to be made public. Personal data or organisations' data stored on
+//! the network is not supposed to be accessed by everyone. Since the data is not published for
+//! everyone, this is called unpublished data. Only the owner(s) shall be able to access and manage
+//! the data based on the permission.
+//!
+//! The network should also be able to provide the flexibility for the users/developers to create
+//! private data which can be versioned or mutable based on their needs.
+//!
+//! ### Private Data
+//!
+//! Users should be able to store private unpublished data on the network which is not accessible by
+//! anyone else.
+//!
+//! ### Shared Data
+//!
+//! Users should be able to store unpublished data on the network and share it with a closed group.
+//! The user should be able to give permissions like read, write, append based on the use case. For
+//! example, a collaborative document which is meant to be worked on within a closed group.
+//!
+//! ## Sequenced and unsequenced data
+//!
+//! Similarly to MutableData, we further sub-divide AppendOnlyData into two distinct sub-categories,
+//! sequenced and unsequenced. For sequenced AppendOnlyData the client must specify the next data
+//! index while appending. For unsequenced AppendOnlyData the client does not have to pass the
+//! index.
+
 use crate::{utils, Error, PublicKey, Result, XorName};
 use multibase::Decodable;
 use serde::{Deserialize, Serialize};
@@ -17,29 +67,44 @@ use std::{
     ops::Range,
 };
 
+/// Published sequenced AppendOnlyData.
 pub type PubSeqAppendOnlyData = SeqAppendOnlyData<PubPermissions>;
+/// Published unsequenced AppendOnlyData.
 pub type PubUnseqAppendOnlyData = UnseqAppendOnlyData<PubPermissions>;
+/// Unpublished sequenced AppendOnlyData.
 pub type UnpubSeqAppendOnlyData = SeqAppendOnlyData<UnpubPermissions>;
+/// Unpublished unsequenced AppendOnlyData.
 pub type UnpubUnseqAppendOnlyData = UnseqAppendOnlyData<UnpubPermissions>;
+/// List of entries.
 pub type Entries = Vec<Entry>;
 
+/// User that can access AppendOnlyData.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, Debug)]
 pub enum User {
+    /// Any user.
     Anyone,
+    /// User identified by its public key.
     Key(PublicKey),
 }
 
+/// An action on AppendOnlyData.
 #[derive(Clone, Copy, Eq, PartialEq)]
 pub enum Action {
+    /// Read from the data.
     Read,
+    /// Append to the data.
     Append,
+    /// Manage permissions.
     ManagePermissions,
 }
 
+/// Index of some data.
 #[derive(Copy, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub enum Index {
-    FromStart(u64), // Absolute index
-    FromEnd(u64),   // Relative index - start counting from the end
+    /// Absolute index.
+    FromStart(u64),
+    /// Relative index - start counting from the end.
+    FromEnd(u64),
 }
 
 impl From<u64> for Index {
@@ -48,7 +113,7 @@ impl From<u64> for Index {
     }
 }
 
-// Set of data, owners, permissions Indices.
+/// Set of data, owners, permissions indices.
 #[derive(Copy, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub struct Indices {
     entries_index: u64,
@@ -57,6 +122,7 @@ pub struct Indices {
 }
 
 impl Indices {
+    /// Constructs a new `Indices`.
     pub fn new(entries_index: u64, owners_index: u64, permissions_index: u64) -> Self {
         Indices {
             entries_index,
@@ -65,27 +131,35 @@ impl Indices {
         }
     }
 
+    /// Returns the last entry index.
     pub fn entries_index(&self) -> u64 {
         self.entries_index
     }
 
+    /// Returns the last owners index.
     pub fn owners_index(&self) -> u64 {
         self.owners_index
     }
 
+    /// Returns the last permissions index.
     pub fn permissions_index(&self) -> u64 {
         self.permissions_index
     }
 }
 
+/// Set of unpublished permissions for a user.
 #[derive(Copy, Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash, Debug)]
 pub struct UnpubPermissionSet {
+    /// `true` if the user can read.
     read: bool,
+    /// `true` if the user can append.
     append: bool,
+    /// `true` if the user can manage permissions.
     manage_permissions: bool,
 }
 
 impl UnpubPermissionSet {
+    /// Constructs a new unpublished permission set.
     pub fn new(read: bool, append: bool, manage_perms: bool) -> Self {
         UnpubPermissionSet {
             read,
@@ -94,12 +168,14 @@ impl UnpubPermissionSet {
         }
     }
 
+    /// Sets permissions.
     pub fn set_perms(&mut self, read: bool, append: bool, manage_perms: bool) {
         self.read = read;
         self.append = append;
         self.manage_permissions = manage_perms;
     }
 
+    /// Returns `true` if `action` is allowed.
     pub fn is_allowed(self, action: Action) -> bool {
         match action {
             Action::Read => self.read,
@@ -109,13 +185,21 @@ impl UnpubPermissionSet {
     }
 }
 
+/// Set of published permissions for a user.
 #[derive(Copy, Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash, Debug)]
 pub struct PubPermissionSet {
+    /// `Some(true)` if the user can append.
+    /// `Some(false)` explicitly denies this permission (even if `Anyone` has required permissions).
+    /// Use permissions for `Anyone` if `None`.
     append: Option<bool>,
+    /// `Some(true)` if the user can manage permissions.
+    /// `Some(false)` explicitly denies this permission (even if `Anyone` has required permissions).
+    /// Use permissions for `Anyone` if `None`.
     manage_permissions: Option<bool>,
 }
 
 impl PubPermissionSet {
+    /// Constructs a new published permission set.
     pub fn new(append: impl Into<Option<bool>>, manage_perms: impl Into<Option<bool>>) -> Self {
         PubPermissionSet {
             append: append.into(),
@@ -123,6 +207,7 @@ impl PubPermissionSet {
         }
     }
 
+    /// Sets permissions.
     pub fn set_perms(
         &mut self,
         append: impl Into<Option<bool>>,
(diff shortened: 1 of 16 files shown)


mrcnski:
Just noticed that `missing_docs` is still allowed on `Request` and `Response`, so still need to work on that.

src/append_only_data.rs
```diff
@@ -109,20 +135,25 @@ impl UnpubPermissionSet {
     }
 }
 
+/// Set of published permissions for a user.
 #[derive(Copy, Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash, Debug)]
 pub struct PubPermissionSet {
+    /// `true` if the user can append.
```
nbaksalyar: Should be
```suggestion
    /// `Some(true)` if the user can append.
```
And probably worth expanding on why it's an `Option` (i.e. the fact that we use permissions for `Anyone` if it's `None`)?

src/append_only_data.rs
```diff
@@ -109,20 +135,25 @@ impl UnpubPermissionSet {
...
     append: Option<bool>,
+    /// `true` if the user can manage permissions.
```
nbaksalyar: Same here, should be
```suggestion
    /// `Some(true)` if the user can manage permissions.
```
And also probably worth mentioning that `Some(false)` _explicitly_ disallows a user to do that (even if `Anyone` has required permissions).

src/append_only_data.rs
```diff
@@ -17,29 +17,44 @@ use std::{
     ops::Range,
 };
 
+/// Published sequenced AppendOnlyData.
```
nbaksalyar, line 70: I think it's also worth adding a small module-level documentation block briefly explaining the differences between the `published` & `unpublished` and `sequenced` & `unsequenced`.

src/append_only_data.rs
```diff
@@ -17,29 +17,44 @@ use std::{
...
+    /// Manage permissions.
     ManagePermissions,
 }
 
+/// Index of some data.
 #[derive(Copy, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub enum Index {
-    FromStart(u64), // Absolute index
-    FromEnd(u64),   // Relative index - start counting from the end
+    /// Absolute index
```
nbaksalyar: Missing a full stop :)
```suggestion
    /// Absolute index.
```

src/append_only_data.rs
```diff
@@ -132,6 +163,7 @@ impl PubPermissionSet {
         self.manage_permissions = manage_perms.into();
     }
 
+    /// Returns `true` if `action` is allowed.
```
nbaksalyar: Returns an `Option<bool>`, so it should be something like
```suggestion
    /// Returns `Some(true)` if `action` is allowed and `Some(false)` if it's not permitted.
    /// `None` would mean that default permissions should be applied.
```

src/append_only_data.rs
```diff
@@ -176,37 +215,44 @@ impl Perm for UnpubPermissions {
...
     }
 
+    /// Gets the complete list of permissions.
     pub fn permissions(&self) -> &BTreeMap<User, PubPermissionSet> {
         &self.permissions
     }
 }
 
 impl Perm for PubPermissions {
+    /// Returns true if `action` is allowed for the provided user.
```
nbaksalyar: This function doesn't return a boolean, so should be
```suggestion
    /// Returns `Ok(())` if `action` is allowed for the provided user and `Err(AccessDenied)` if
    /// this action is not permitted.
```

src/append_only_data.rs
```diff
@@ -271,76 +329,77 @@ struct AppendOnly<P: Perm> {
...
+    /// Fetches owner at index.
     fn owner(&self, owners_index: impl Into<Index>) -> Option<&Owner>;
 
-    /// Get a complete list of owners from the entry in the permissions list at the specified index.
+    /// Gets a complete list of owners from the entry in the permissions list at the specified
+    /// index.
     fn owners_range(&self, start: Index, end: Index) -> Option<&[Owner]>;
 
-    /// Add a new owner entry.
+    /// Adds a new owner entry.
```
nbaksalyar, line 448: Worth expanding on what `owners_index` means in this context (should be equal to `current owners index + 1`)
Something like
```rust
/// If the specified `owners_index` does not match the last recorded owners index + 1, an
/// error will be returned.
```

src/append_only_data.rs
```diff
@@ -271,76 +329,77 @@ struct AppendOnly<P: Perm> {
...
     fn permissions_index(&self) -> u64;
 
-    /// Get a complete list of permissions from the entry in the permissions list at the specified
+    /// Gets a complete list of permissions from the entry in the permissions list at the specified
     /// index.
     fn permissions_range(&self, start: Index, end: Index) -> Option<&[P]>;
 
-    /// Add a new permissions entry.
+    /// Adds a new permissions entry.
     /// The `Perm` struct should contain valid indices.
```
nbaksalyar, line 432: Also worth expanding on what `permissions_index` means in this context (should be equal to `current permissions index + 1`)

src/append_only_data.rs
```diff
@@ -271,76 +329,77 @@ struct AppendOnly<P: Perm> {
...
+    /// Returns the last owners index.
     fn owners_index(&self) -> u64;
 
-    /// Return the last permissions index.
+    /// Returns the last permissions index.
     fn permissions_index(&self) -> u64;
 
-    /// Get a complete list of permissions from the entry in the permissions list at the specified
+    /// Gets a complete list of permissions from the entry in the permissions list at the specified
     /// index.
```
nbaksalyar: should be
```suggestion
     /// indices.
```
?

src/append_only_data.rs
```diff
@@ -431,15 +498,18 @@ macro_rules! impl_appendable_data {
...
                 self.inner.data.last()
             }
 
+            /// Gets a complete list of permissions.
             fn permissions(&self, index: impl Into<Index>) -> Option<&P> {
                 let index = to_absolute_index(index.into(), self.inner.permissions.len())?;
                 self.inner.permissions.get(index)
             }
 
+            /// Returns the owner key.
```
nbaksalyar: It's not entirely clear what `key` means in this context, I'd suggest replacing it with something like
```suggestion
            /// Returns the owner's public key and the indices at the time it was added.
```

src/append_only_data.rs
```diff
@@ -808,34 +909,62 @@ impl Data {
...
     pub fn is_seq(&self) -> bool {
         self.kind().is_seq()
     }
 
+    /// Returns `true` if unsequenced.
     pub fn is_unseq(&self) -> bool {
         self.kind().is_unseq()
     }
 
+    /// Checks permissions for given `action` for the provided user.
```
nbaksalyar, line 1016: Needs to be expanded to describe which error types we return here and in which cases.

src/append_only_data.rs
```diff
@@ -881,6 +1014,8 @@ impl Data {
         }
     }
 
+    /// Returns a tuple containing the last entries index, last owners index, and last permissions
```
nbaksalyar, line 1093: We need to explain why it is a `Result` and in which case an error might be returned.

src/append_only_data.rs
```diff
@@ -908,6 +1045,7 @@ impl Data {
         }
     }
 
+    /// Checks if the requester is the last owner.
```
nbaksalyar, line 1126: We need to explain why it is a Result and in which case an error might be returned.

src/lib.rs
```diff
@@ -158,6 +164,7 @@ impl From<AData> for Data {
     Copy, Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize, Default, Debug,
 )]
 pub struct AppPermissions {
+    /// Whether this app has permissions to transfer coins.
```
nbaksalyar, line 167: Could you please change the doc above from
```
/// Permissions for an app stored by the Elders.
```
to
```
/// Permissions for an app stored by the Client Handlers.
```
for clarity?

src/lib.rs
```diff
@@ -210,21 +219,31 @@ impl Distribution<XorName> for Standard {
 #[allow(clippy::large_enum_variant)]
 #[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
 pub enum Message {
+    /// Request with the message ID.
     Request {
+        /// Request.
         request: Request,
+        /// Associated message ID.
         message_id: MessageId,
+        /// Optional signature.
```
nbaksalyar: Why is it optional? Should be expanded into something like
```suggestion
        /// Signature of `(request, message_id)`. Optional if the request is read-only.
```

src/lib.rs
```diff
@@ -259,7 +278,9 @@ impl Default for MessageId {
 #[allow(clippy::large_enum_variant)]
 #[derive(Serialize, Deserialize)]
 pub enum Challenge {
+    /// Request containing the challenge.
```
nbaksalyar: I'd suggest adding some more info
```suggestion
    /// Request from the Vault to clients containing the challenge.
```

src/lib.rs
```diff
@@ -259,7 +278,9 @@ impl Default for MessageId {
...
     Request(PublicId, Vec<u8>),
+    /// Response.
```
nbaksalyar: Same here:
```suggestion
    /// Response from clients to Vaults, containing their public ID and the challenge
    /// signature created using their corresponding secret key.
```

src/immutable_data.rs
```diff
@@ -17,12 +17,14 @@ use std::{
 };
 use tiny_keccak;
 
-/// Maximum allowed size for a serialised Immutable Data (ID) to grow to
+/// Maximum allowed size for a serialised ImmutableData to grow to.
 pub const MAX_IMMUTABLE_DATA_SIZE_IN_BYTES: u64 = 1024 * 1024 + 10 * 1024;
 
+/// Unpublished ImmutableData: an immutable chunk of data which can be deleted.
```
nbaksalyar: Probably worth adding that it also can't be fetched by anyone but the listed `owner`.

src/mutable_data.rs
```diff
@@ -166,61 +172,63 @@ pub enum Action {
...
             pub fn permissions(&self) -> BTreeMap<PublicKey, PermissionSet> {
                 self.permissions.clone()
             }
 
+            /// Gets the permissions for the provided user.
             pub fn user_permissions(&self, user: PublicKey) -> Result<&PermissionSet> {
                 self.permissions.get(&user).ok_or(Error::NoSuchKey)
             }
 
+            /// Checks if the provided user is an owner.
```
nbaksalyar, line 251:
```suggestion
            /// Checks if the provided user is an owner.
            /// Returns `Ok(())` on success and `Err(Error::AccessDenied)` if the user is not an owner.
```

src/mutable_data.rs
```diff
@@ -229,6 +237,7 @@ macro_rules! impl_mutable_data {
                 }
             }
 
+            /// Checks permissions for given `action` for the provided user.
```
nbaksalyar, line 263:
```suggestion
            /// Checks permissions for given `action` for the provided user.
            /// Returns `Err(Error::AccessDenied)` if the permission check has failed.
```

src/mutable_data.rs
```diff
@@ -244,7 +253,7 @@ macro_rules! impl_mutable_data {
                 }
             }
 
-            /// Insert or update permissions for the provided user.
+            /// Inserts or updates permissions for the provided user.
```
nbaksalyar, line 281: We should mention what the `version` parameter means and that it must be equal to `current version + 1`.

src/mutable_data.rs
```diff
@@ -259,7 +268,7 @@ macro_rules! impl_mutable_data {
                 Ok(())
             }
 
-            /// Delete permissions for the provided user.
+            /// Deletes permissions for the provided user.
```
nbaksalyar, line 301: We should mention what the version parameter means and that it must be equal to current version + 1.

src/mutable_data.rs
```diff
@@ -272,7 +281,7 @@ macro_rules! impl_mutable_data {
                 Ok(())
             }
 
-            /// Delete user permissions without performing any validation.
+            /// Deletes user permissions without performing any validation.
```
nbaksalyar, line 319: Same comment about version.

src/mutable_data.rs
```diff
@@ -286,7 +295,7 @@ macro_rules! impl_mutable_data {
                 true
             }
 
-            /// Change owner of the mutable data.
+            /// Changes the owner.
```
nbaksalyar, line 338: Same comment about version.

src/mutable_data.rs
```diff
@@ -296,7 +305,7 @@ macro_rules! impl_mutable_data {
                 Ok(())
             }
 
-            /// Change the owner without performing any validation.
+            /// Changes the owner without performing any validation.
```
nbaksalyar, line 353: Same comment about version.

src/mutable_data.rs
```diff
@@ -354,26 +363,27 @@ impl UnseqMutableData {
...
         &self.data
     }
 
-    /// Removes and returns all entries
+    /// Removes and returns all entries.
     pub fn take_entries(&mut self) -> UnseqEntries {
         mem::replace(&mut self.data, BTreeMap::new())
     }
 
+    /// Mutates entries based on `actions` for the provided user.
```
nbaksalyar, line 435: Please add that it returns `Err(InvalidEntryActions)` in case if the mutation parameters are invalid.

src/mutable_data.rs
```diff
@@ -505,7 +515,7 @@ impl SeqMutableData {
         mem::replace(&mut self.data, BTreeMap::new())
     }
 
-    /// Mutates entries (key + value pairs) in bulk
+    /// Mutates entries (key + value pairs) in bulk.
```
nbaksalyar, line 569: Same comment about the possible errors that this function might return.

src/mutable_data.rs
```diff
@@ -816,43 +869,46 @@ impl From<UnseqMutableData> for Data {
...
         Default::default()
     }
 
+    /// Gets the actions.
     pub fn actions(&self) -> &BTreeMap<Vec<u8>, SeqEntryAction> {
         &self.actions
     }
 
-    /// Insert a new key-value pair
+    /// Inserts a new key-value pair.
```
nbaksalyar, line 964: It's also worth explaining what the `version` parameter means and what it should be equal to.

src/mutable_data.rs
```diff
@@ -864,7 +920,7 @@ impl SeqEntryActions {
         self
     }
 
-    /// Update an existing key-value pair
+    /// Updates an existing key-value pair.
```
nbaksalyar, line 979: It's also worth explaining what the `version` parameter means and what it should be equal to.

src/mutable_data.rs
```diff
@@ -876,7 +932,7 @@ impl SeqEntryActions {
         self
     }
 
-    /// Delete an entry
+    /// Deletes an entry
```
nbaksalyar: Missing full stop:
```suggestion
    /// Deletes an entry.
```
It's also worth explaining what the `version` parameter means and what it should be equal to.

src/mutable_data.rs
```diff
@@ -895,13 +951,14 @@ impl From<BTreeMap<Vec<u8>, SeqEntryAction>> for SeqEntryActions {
...
 
+/// Unsequenced Entry Actions for given entry keys.
 #[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize, Debug, Default)]
 pub struct UnseqEntryActions {
     actions: BTreeMap<Vec<u8>, UnseqEntryAction>,
 }
 
 impl UnseqEntryActions {
-    /// Create a new Unsequenced Entry Actions list
+    /// Creates a new Unsequenced Entry Actions list
```
nbaksalyar: Missing full stop:
```suggestion
    /// Creates a new Unsequenced Entry Actions list.
```

src/public_key.rs
```diff
@@ -19,15 +19,19 @@ use std::{
...
+    /// Ed25519 public key.
     Ed25519(ed25519_dalek::PublicKey),
+    /// BLS public key.
     Bls(threshold_crypto::PublicKey),
+    /// BLS public key share.
     BlsShare(threshold_crypto::PublicKeyShare),
 }
 
 impl PublicKey {
+    /// Returns `true` if `signature` matches the message.
```
nbaksalyar: This function does not return a boolean.
Should be something like
```rust
/// Returns `Ok(())` if `signature` matches the message and `Err(Error::InvalidSignature)` otherwise.
```

src/append_only_data.rs
mrcnski: Is it intentional that this always returns `Ok`? Can I fix it now or should I write an issue?

nbaksalyar: Good catch :+1: But yes, let's record it as an issue and leave the API as it is for now.

mrcnski: Looks like Adam's PR already addressed this :)

PR #97, opened 2019-08-05
feat/mdata: impl helpers for MData SeqEntryActions

Yoga07:
- Implements `From` trait for SeqEntryActions (from `BTreeMap<Vec<u8>,SeqEntryAction>`)
- Implements getter function for `actions` field in SeqEntryActions

### DIFF
--- a/src/mutable_data.rs
+++ b/src/mutable_data.rs
@@ -839,6 +839,10 @@ impl SeqEntryActions {
         Default::default()
     }
 
+    pub fn actions(&self) -> &BTreeMap<Vec<u8>, SeqEntryAction> {
+        &self.actions
+    }
+
     /// Insert a new key-value pair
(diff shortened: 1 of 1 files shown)


src/mutable_data.rs
```diff
@@ -839,6 +839,10 @@ impl SeqEntryActions {
         Default::default()
     }
 
+    pub fn actions(&self) -> BTreeMap<Vec<u8>, SeqEntryAction> {
+        self.actions.clone()
```
nbaksalyar: Maybe better to return a ref here and make a user choose to whether they want to clone it or not? i.e.
```rust
pub fn actions(&self) -> &BTreeMap<Vec<u8>, SeqEntryAction> {
    &self.actions
}
```

PR #101, opened 2019-08-07
feat/kind: add `from_flag` method to `Kind` types

m-cat:


src/append_only_data.rs
```diff
@@ -671,6 +671,21 @@ pub enum Kind {
 }
 
 impl Kind {
+    /// Creates `Kind` from `published` and `sequenced` flags.
+    pub fn from_flags(published: bool, sequenced: bool) -> Self {
+        if published {
```
nbaksalyar: I would suggest rewriting it using `match`:
```rust
match (published, sequenced) {
  (true, true) => Kind::PubSeq,
  (true, false) => Kind::PubUnseq,
  (false, true) => Kind::UnpubSeq,
  (false, false) => Kind::UnpubUnseq,
}
```
but it's a matter of style preference :)

mrcnski: That looks cleaner, will change it.

PR #105, opened 2019-08-12
feat/adata: add missing AData methods

m-cat:


src/mutable_data.rs
```diff
@@ -1016,11 +1016,12 @@ impl From<BTreeMap<Vec<u8>, SeqEntryAction>> for SeqEntryActions {
 /// Unsequenced Entry Actions for given entry keys.
 #[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, Serialize, Deserialize, Debug, Default)]
 pub struct UnseqEntryActions {
-    actions: BTreeMap<Vec<u8>, UnseqEntryAction>,
+    /// A BTreeMap containing keys to which the corresponding action is to be performed.
+    pub actions: BTreeMap<Vec<u8>, UnseqEntryAction>,
```
lionel-faber: This is changing in #107 so we don't need to change this here.

mrcnski: Yeah, this change will just disappear after rebasing, but I figure the fixes to the documentation inconsistencies are still worth making.

lionel-faber: Of course, yes. I'm happy with this as is.

PR #107, opened 2019-08-12
feat/mdata: add getter functions and add_action for MData Seq/Unseq EntryActions

Yoga07:


src/mutable_data.rs
```diff
@@ -961,6 +962,16 @@ impl SeqEntryActions {
...
     }
 
+    /// Converts the SeqEntryActions into a BTreeMap containing keys with their corresponding action
+    pub fn into_actions(self) -> BTreeMap<Vec<u8>, SeqEntryAction> {
+        self.actions
+    }
+
+    /// Gets the actions mutably
+    pub fn get_mut(&mut self) -> &mut BTreeMap<Vec<u8>, SeqEntryAction> {
+        self.actions.borrow_mut()
```
mrcnski: Maybe I'm wrong but can't this simply be 

```rust
&mut self.actions
```

mrcnski: One more comment: since already have a method called `actions`, this should be called `actions_mut` for consistency.

mrcnski:
I'm wondering if it's really worthwhile to have three separate methods just to access the internal data. It might be better to make the internal field public:

```rust
pub struct SeqEntryActions(pub BTreeMap<Vec<u8>, SeqEntryAction>)
```

what do you think?

Yoga07:
Yea, I did think of making it public some time back. But, are there chances of malicious actors getting their hands on somebody's entry_actions and tampering them? Not sure if I am putting it the right way. :)

mrcnski:
I don't think it would be a problem unless the malicious actor could push the changes to the network. And `get_mut` would make them mutable anyway :)

Pinging @lionel1704 for his thoughts.

lionel-faber:
> the malicious actor could push the changes to the network

@m-cat is right here. Changing it would be of no use unless the change could be pushed to the network. Even if the entry actions are updated _after_ the request is sent if the entry actions change then the signature validation would fail.

Yoga07:
Yup! Makes sense :) Making it `pub`.

mrcnski:
Do you think we can do the same change for `UnseqEntryActions` while we're at it?

Yoga07:
I am actually thinking if we need a separate PR for this? Can we merge this one with https://github.com/maidsafe/safe-nd/pull/105/? :)

src/mutable_data.rs
```diff
@@ -996,28 +990,26 @@ impl SeqEntryActions {
...
 }
 
 impl From<SeqEntryActions> for BTreeMap<Vec<u8>, SeqEntryAction> {
     fn from(actions: SeqEntryActions) -> Self {
-        actions.actions
+        actions.0
     }
 }
 
 impl From<BTreeMap<Vec<u8>, SeqEntryAction>> for SeqEntryActions {
```
mrcnski: It looks like a `From` impl similar to this one is missing for `UnseqEntryActions`, can you please add it?

PR #109, opened 2019-08-20
Validate permissions correctly for AData

lionel1704:


src/identity/app.rs
```diff
@@ -128,6 +128,6 @@ impl Debug for PublicId {
 impl Display for PublicId {
     #[allow(trivial_casts)]
     fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
-        (self as &Debug).fmt(formatter)
+        (self as &dyn Debug).fmt(formatter)
```
mrcnski: Looks like Adam fixed this in his PR in a more concise way.

Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "MIT OR BSD-3-Clause"
 name = "safe-nd"
 readme = "README.md"
 repository = "https://github.com/maidsafe/safe-nd"
-version = "0.2.1"
+version = "0.2.2"
```
mrcnski: We need to update `Changelog.md` as well. Also, version changes usually happen in their own PRs (i.e. without any other changes).

PR #110, opened 2019-08-20
Add refund reason to `Request::TransferCoins`

madadam:


src/request.rs
```diff
@@ -220,6 +220,8 @@ pub enum Request {
         amount: Coins,
         /// The ID of the transaction.
         transaction_id: TransactionId,
+        /// If this transfer is a refund, this field contains the reason for it. Otherwise it is `None`.
```
nbaksalyar, line 223: I would probably add just one bit saying something like "it should be always set to `None` for requests sent by clients"

src/identity/app.rs
```diff
@@ -128,6 +128,6 @@ impl Debug for PublicId {
 impl Display for PublicId {
     #[allow(trivial_casts)]
```
mrcnski, line 129: Can we get rid of these `allow`s now?

nbaksalyar:
We decided to add this field to the [vault-level RPC struct](https://github.com/maidsafe/safe_vault/blob/87f53f9aba4a684c70609174d672672bc5543253/src/rpc.rs#L17) instead.

PR #116, opened 2019-09-11
refactor/permission: granularize AppPermissions struct

Yoga07:
Required for maidsafe/safe_client_libs#989 and maidsafe/safe_client_libs#990
- Adds `allow_mutations` and `get_balance` booleans to `AppPermissions`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -166,6 +166,10 @@ impl From<AData> for Data {
 pub struct AppPermissions {
     /// Whether this app has permissions to transfer coins.
     pub transfer_coins: bool,
+    /// Whether this app has permissions to perform mutations.
+    pub perform_mutations: bool,
+    /// Whether this app has permissions to read the coin balance.
+    pub get_balance: bool,
 }
 
 /// Constant byte length of `XorName`.


src/lib.rs
```diff
@@ -166,6 +166,10 @@ impl From<AData> for Data {
 pub struct AppPermissions {
     /// Whether this app has permissions to transfer coins.
     pub transfer_coins: bool,
+    /// Whether this app has permissions to perform mutations.
+    pub allow_mutations: bool,
+    /// Whether this app has permissions to read balances.
```
nbaksalyar: Probably should be
```suggestion
    /// Whether this app has permissions to read the coin balance.
```

src/lib.rs
```diff
@@ -166,6 +166,10 @@ impl From<AData> for Data {
 pub struct AppPermissions {
     /// Whether this app has permissions to transfer coins.
     pub transfer_coins: bool,
+    /// Whether this app has permissions to perform mutations.
+    pub allow_mutations: bool,
```
mrcnski: I wonder if `allow_mutations` is the best name for this permission. "allow" is redundant, e.g. the `transfer_coins` permission is not `allow_transfer_coins`. Also, since the other two permissions are descriptions of the action being allowed, I'm thinking this should be `perform_mutation`. What do you think?

nbaksalyar: Agreed, `perform_mutations` sounds better :+1:

Issue #119, opened 2019-10-21
Add missing Get function for AppendOnlyData.


oetyng:
`Owners` and `Permissions` can be fetched by their index, but an `Entry` (the data itself) cannot, which would be useful, instead of trying to use `get_in_range`. It seems get by index is just missing here.

Proposing to add

(in `trait AppendOnlyData<P>` in `append_only_data.rs`)

    /// Fetches entry at index.
    fn entry(&self, entry_index: impl Into<Index>) -> Option<&Entry>;

as well as

(in `impl<P> AppendOnlyData<P>` in `append_only_data.rs`)

    /// Returns the entry at the index.
    fn entry(&self, entry_index: impl Into<Index>) -> Option<&Entry> {
        let index = to_absolute_index(entry_index.into(), self.inner.data.len())?;
        self.inner.data.get(index)
    }

and

(in `impl Data` in `append_only_data.rs`)

    /// Fetches entry at index.
    pub fn entry(&self, entry_index: impl Into<Index>) -> Option<&Entry> {
        match self {
            Data::PubSeq(data) => data.entry(entry_index),
            Data::PubUnseq(data) => data.entry(entry_index),
            Data::UnpubSeq(data) => data.entry(entry_index),
            Data::UnpubUnseq(data) => data.entry(entry_index),
        }
    }

as well as corresponding test.

Any other places needing change for this @m-cat ?

mrcnski:
This change sounds good to me 👍 

> Any other places needing change for this @m-cat ?

I think that's all of them

S-Coyle:
Closing - data type replaced with `Sequence` and `Map`

PR #121, opened 2019-10-21
fix/get: add missing api to AppendOnlyData trait

oetyng:
Getting Entry by index is missing, but exists for Owner and Permissions.
REFERENCE: #119

### DIFF
--- a/src/append_only_data.rs
+++ b/src/append_only_data.rs
@@ -1370,81 +1370,42 @@ mod tests {
 
     #[test]
     fn get_entry() {
-        // pub, unseq
-        let mut data = PubUnseqAppendOnlyData::new(rand::random(), 10);
+        let name: XorName = rand::random();
+        let tag = 10;
+
+        let mut pub_seq = PubSeqAppendOnlyData::new(name, tag);
+        let mut unpub_seq = UnpubSeqAppendOnlyData::new(name, tag);
+        let mut pub_unseq = PubUnseqAppendOnlyData::new(name, tag);
+        let mut unpub_unseq = UnpubUnseqAppendOnlyData::new(name, tag);
+
         let entries = vec![
             Entry::new(b"key0".to_vec(), b"value0".to_vec()),
             Entry::new(b"key1".to_vec(), b"value1".to_vec()),
         ];
-        unwrap!(data.append(entries));
-        let data = Data::from(data);
 
-        assert_eq!(
-            data.entry(Index::FromStart(0)),
-            Some(&Entry::new(b"key0".to_vec(), b"value0".to_vec()))
-        );
-        assert_eq!(
-            data.entry(Index::FromStart(1)),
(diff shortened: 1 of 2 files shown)


oetyng:
Hmm. It is failing Travis due to code that did not change in this PR. Not sure why.

Yoga07:
> Hmm. It is failing Travis due to code that did not change in this PR. Not sure why.

It's from here @oetyng.
https://github.com/maidsafe/safe-nd/pull/121/files#diff-d03d8576cec38966d9ac8eda6c082f23R1432
Running `rustfmt` would clean it up :)

mrcnski (dismissed):

oetyng:
Dismissed by push, sorry @m-cat :)
Now tests are cleaner though.

mrcnski (dismissed):
Looks great 😄

oetyng:
Oh not again, I thought I would make it before you reviewed!
I improved it one slight tiny bit more :)

src/append_only_data.rs
```diff
@@ -1398,11 +1398,11 @@ mod tests {
         for data in data_vec {
             assert_eq!(
                 data.entry(Index::FromStart(0)),
-                Some(&Entry::new(b"key0".to_vec(), b"value0".to_vec()))
+                Some(&entries.clone()[0])
```
mrcnski: Won't this clone the entire vec? I think `entries[0].clone()` is what you want, but I haven't tested it.

Of course, this is just a test so the performance here isn't important.

mrcnski:
> Oh not again, I thought I would make it before you reviewed!
> I improved it one slight tiny bit more :)

No worries 🙂 

Looks good, but I'll wait this time until the CI issues are fixed

src/append_only_data.rs
oetyng: You're right! And actually, no need for clone at all there I realised. I was first thinking that the cloned entries vec was a shallow clone, and that the entries within needed to be cloned. But apparently cloning the vec must have cloned the contents as well?

mrcnski: If the entries inside the `Vec` are a `Clone` type, then they will be cloned (even if they are a reference), otherwise if they are a reference but not a `Clone` type the references themselves will be cloned (shallow clone), otherwise you get a compiler error

lionel-faber:
These changes look good. CI is failing because of rust-lang/rust-clippy#4326.
@m-cat has a temporary fix in #123 so once that is merged, we can rebase merge this PR.
Thanks @oetyng!

mrcnski:
@oetyng Can you include this change in this PR:

https://github.com/maidsafe/safe-nd/pull/123/files#diff-d03d8576cec38966d9ac8eda6c082f23R60

so we can get this merged quicker without waiting on #123 (not sure when that will be reviewed/approved and its a larger change)

lionel-faber:
Closing this since this data type is being replaced with `Sequence` and `Map`

PR #123, opened 2019-11-01
feat/keys: add `Keypair` and remove bls methods for full IDs

mrcnski:
GitHub doesn't show a nice diff for the `keys.rs` file since it was renamed from `public_key.rs`. Reviewers might want to generate a diff locally.


src/keys.rs
```diff
@@ -0,0 +1,371 @@
...
+}
+
+impl SecretKey {
+    /// Generates a random BLS key.
+    #[cfg(test)]
+    pub fn random_bls() -> Self {
+        Self::Bls(threshold_crypto::SecretKey::random())
+    }
+
+    // TODO: implement zbase32 encoding and decoding?
```
lionel-faber: We might need this since a user may like to encode his secret key and store it as a string somewhere locally.

lionel-faber (commented):
The changes look good.  Seems like CI is failing since using `Self::` in enums is experimental.

src/identity/mod.rs
```diff
@@ -37,9 +33,45 @@ impl PublicId {
...
+            Some(id)
+        } else {
+            None
+        }
+    }
+
+    /// Returns the entity's public key, if applicable.
+    pub fn public_key(&self) -> Option<PublicKey> {
+        match self {
+            Self::Node(pub_id) => pub_id.bls_public_key().map(|key| key.into()),
```
lionel-faber: Do we specifically need the BLS public key?

lionel-faber (commented):
Looks good. Just one comment.

src/identity/mod.rs
mrcnski: What should be returned here? The ed25519 key?

lionel-faber: Yes I believe so. The vault by default doesn't have a BLS key, it generates only the ed25519 key and that's the key that represents the Vault's identity.

src/identity/mod.rs
```diff
@@ -37,9 +33,45 @@ impl PublicId {
...
+    pub fn app_public_id(&self) -> Option<&app::PublicId> {
+        if let Self::App(id) = self {
+            Some(id)
+        } else {
+            None
+        }
+    }
+
+    /// Returns the entity's public key, if applicable.
+    pub fn public_key(&self) -> Option<PublicKey> {
```
lionel-faber: Can this function return just `PublicKey` now?
(Clippy has also failed)

mrcnski: The failures on Jenkins are due to an out of date container. However, I think @jacderida said that the container fails to build -- seems like it can't be updated without the lint allowance in this PR. I would say let's ignore Jenkins for this PR.

lionel-faber: That's cool then! :+1:

lionel-faber (dismissed):
Looks good! Thanks! :+1:

lionel-faber (approved):

PR #126, opened 2019-11-15
AD indices refactor

### LINKED ISSUE #120: Adjust naming and docs to reflect actual index behaviour
An entry is specified as a struct with key and value.

    pub struct Entry {
        /// Key.
        pub key: Vec<u8>,
        /// Contained data.
        pub value: Vec<u8>,
    }

Then follows this:

    /// Returns the last entries index.
    fn entries_index(&self) -> u64 {
        self.inner.data.len() as u64
     }

The comment and function name are misleading. It is returning the next unused index, not the "last entries index".
Consider a data with 0 items. There is no entry in it, and so it is not returning the last entries index.
A data with 1 item, will return 1, which is not the index of the last entry. The last entry, is the one and only existing Entry, at index 0.

Other examples of this:

    /// If the specified last_entries_index does not match the last recorded entries index, an
    /// error will be returned.
    fn append(&mut self, mut entries: Entries, last_entries_index: u64) -> Result<()> {
    [..]
        if last_entries_index != self.inner.data.len() as u64 {
            return Err(Error::InvalidSuccessor(self.inner.data.len() as u64));
    }

This inconsistency percolates all the way to the top layers of the API.

------------------------

Proposed change is to acknowledge (in all code and names everywhere) that we are actually passing in _next unused index_. Then passing in `0` when expecting an empty AD is exactly as it should be.

Name change in code and documentation:

`last_entries_index` -> `next_unused_index`

oetyng:
Use unambiguous and consistent name referencing of indices.

Closes: #120 

This introduces an `AppendOnlyData` indices naming convention, as to avoid confusion and simplify in all repos and on all levels. Applies to `Entries`, `Owners` and `Permissions`.
The goal is to never mix-up what current, next, last index means or refers to, because it does lead to bugs and issues (like _off by one error_).

## Index naming convention

- `Current` -> the index where the last item in the collection is. Can be `None` when empty. 
--> `Current item` resides at `current index`.
- `Expected` -> the index that is expected to be current when adding an item, the index where additions are expected. 
--> A new item to be appended, will then reside at what is now the `expected index`. After it is appended, it is the `current item` at the `current index`.

Not yet found use case for `Previous` and `Next`, but they are part of the convention in the following way:
From any index we refer to preceding and subsequent index as `Previous` and `Next` respectively.
From index 0, `Previous` is `None`. From `Current` index, `Next` is `None`.

IndicesConcept


src/append_only_data.rs
```diff
@@ -581,7 +596,7 @@ macro_rules! impl_appendable_data {
                 self.inner.permissions.get(index)
             }
 
-            /// Returns the owner's public key and the indices at the time it was added.
+            /// Returns the owner's public key and the indices at the time it was added. // <--- confusing docs
```
mrcnski: I'm not sure if you meant to commit this, but won't this comment appear in the public documentation for this function? I would say either move it into a separate `TODO:` comment or just address it now.

mrcnski (changes requested):
Looks great!

I just noticed that the documentation in `append_only_data.rs` is not generated by `cargo doc`, because it is a private module! I'll write an issue about that.

src/append_only_data.rs
oetyng: I did not actually, I'm sorry it slipped through. It was a note to myself to get to the bottom with what it meant. It can be removed.

mrcnski (dismissed):

src/append_only_data.rs
```diff
@@ -610,35 +625,37 @@ macro_rules! impl_appendable_data {
...
 
             /// Adds a new owner entry.
             ///
-            /// If the specified `owners_index` does not match the last recorded owners index + 1,
-            /// an error will be returned.
-            fn append_owner(&mut self, owner: Owner, owners_index: u64) -> Result<()> {
-                if owner.entries_index != self.entries_index() {
-                    return Err(Error::InvalidSuccessor(self.entries_index()));
+            /// If the specified `expected_index` does not equal
+            /// the count of owners index, an error will be returned.
```
lionel-faber:
```suggestion
            /// the count of owners, an error will be returned.
```
Do you think we should change this as well?

lionel-faber (commented):
Great work with this @oetyng !
Just one small comment.

src/append_only_data.rs
oetyng: Absolutely, I'll do the change.

mrcnski (approved):

lionel-faber:
Closing this since this data type is being replaced with `Sequence` and `Map`

PR #128, opened 2019-11-18
chore/ci: deprecate Jenkins in favor of GitHub Actions

mrcnski:
Closes https://github.com/maidsafe/QA/issues/109


.github/workflows/master.yml
```diff
@@ -0,0 +1,36 @@
+# Push to master workflow.
+#
+# Less strict than the PR workflow because checks have run already.
+
+name: Master
+
+on:
```
lionel-faber, line 10: Can we can change this to:
```
on:
 push:
  branches:
   - master
```
Since we don't push directly to the master branch.

(Unless a PR merge doesn't trigger the push event)

.github/workflows/master.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  test:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+    steps:
+      - uses: actions/checkout@v1
+        if: github.event.pull_request.merged
```
lionel-faber: The conditional can be moved to the Job level as well. It can be placed anywhere before `steps:`
That way we don't need to have the same conditional for every step

[Here's an example](https://github.com/lionel1704/gh-actions-demo/blob/d552e579abf47824756cb39bde96facaa8198c31/.github/workflows/releases.yml#L7)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+#
+# Runs full suite of checks, with warnings treated as errors.
+
+name: PR
+
+on: pull_request
+
+env:
+  RUST_BACKTRACE: 1
+  RUST_FLAGS: "-D warnings"
```
lionel-faber: *RUSTFLAGS
:)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions-rs/cargo@v1
+        if: github.event.pull_request.merged
+        with:
+          command: package
+          args: --manifest-path=safe_core/Cargo.toml
+      - uses: actions-rs/cargo@v1
+        if: github.event.pull_request.merged
+        with:
+          command: publish
+          args: --manifest-path=safe_core/Cargo.toml --dry-run
```
lionel-faber: We don't need the `--manifest-path=safe_core/Cargo.toml` here :smile: 
Also why is this just a dry run?

lionel-faber (changes requested):
Nice work! Just a few comments.

I like the separate workflows for PRs and a lighter workflow on master. One question: Instead of separating master and publish workflows is it any better / faster to include it as a job after running tests on master? Or does `cargo publish` build everything from scratch without re-using anything from the older `cargo test` builds.

It'd be nice to confirm that we don't change the `CHANGELOG.md` file unless it's a version change commit. We might keep adding new features to the changelog under a "Unreleased" heading until a version change PR is raised, so we'll have to make sure of our process there.

.github/workflows/master.yml
mrcnski: I'm not sure I know what you mean @lionel1704. And yeah, I think PR merges don't trigger a push event.

lionel-faber: I just checked on my repo. PR merges are triggering the push event on the master branch.
So we don't actually need the conditional for merged pull requests. Instead we can trigger this workflow using the `push` event. Does that make sense?

mrcnski: Ah that's great, will make the change! I guess a lot of the info I've been reading is out of date.

.github/workflows/master.yml
```diff
@@ -0,0 +1,63 @@
...
+      - uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --verbose --release
+      # Get information for the publish.
+      - shell: bash
+        id: commit_message
+        run: |
+          commit_message=$(git log --format=%B -n 1 ${{ github.sha }})
+          echo "::set-output name=commit_message::$commit_message"
```
jacderida: I think you need to include this in the `publish` job. You can't reference information from steps in other jobs. I ran into this yesterday with SCL.

It also makes me wonder if the condition at the job level will work, since you won't have the commit message before the job starts. I will be able to verify this with SCL.

.github/workflows/master.yml
```diff
@@ -0,0 +1,63 @@
...
+ push:
+  branches:
+   - master
+
+env:
+  RUST_BACKTRACE: 1
+
+jobs:
+  build:
+    name: Build
```
jacderida, line 20: Due to the fact that safe-nd doesn't need any build artifacts because we don't deploy anything, and the publish also doesn't need to actually run a build, technically we don't actually even need a build here, unless we just want to verify that it builds. If you were satisfied that was the case in the PR, we would be all set. However, it doesn't really do much harm to do it either.

mrcnski: > unless we just want to verify that it builds.

Yeah, basically I wanted to have a README badge that reports the latest build status. This way people who check out the latest code in master will know what to expect when they try to build it locally.

Although this currently doesn't make much sense unless we merge code that doesn't build. For this to be more useful, we could also make this workflow run on a schedule -- this way we can stay on top of things breaking due to e.g. new rustc version. What do you think?

lionel-faber (dismissed):

mrcnski:
This isn't ready yet -- trying to figure out how to make sure the Publish step doesn't run on the cron schedule.

lionel-faber (approved):

PR #130, opened 2019-11-25
Add functions to convert a signature into bytes and vice versa

lionel-faber:


src/keys.rs
```diff
@@ -192,6 +195,47 @@ pub enum Signature {
...
+        let (mut bytes, variant) = match self {
+            Self::Bls(key) => (key.to_bytes().to_vec(), [0u8]),
+            Self::BlsShare(key) => (key.to_bytes().to_vec(), [1u8]),
+            Self::Ed25519(key) => (key.to_bytes().to_vec(), [2u8]),
+        };
+        bytes.extend_from_slice(&variant);
+        bytes
+    }
+
+    /// Returns the Signature from it's byte represenation
```
mrcnski, line 211: Typos: should be "its" and "representation"

src/keys.rs
```diff
@@ -192,6 +195,47 @@ pub enum Signature {
...
+    pub fn from_bytes(mut bytes: Vec<u8>) -> Result<Self> {
```
mrcnski, line 212: Can this parameter take `&[u8]` instead? It's more general because you can pass a reference to a `Vec<u8>` and it will deref. From what I've seen in the std library and the ecosystem `from_bytes` usually takes `&[u8]`. Also, does this have to be mutable?

lionel-faber:
Just realized that we might not even need these functions as Serialize and Deserialize are already derived.

mrcnski (changes requested):

lionel-faber:
These changes are not really required. The `Signature` type has the serialisation traits already derived.

PR #133, opened 2019-12-10
Impl From trait for client::FullId

lionel-faber:
- implement the From trait for client::FullId to allow conversion from
the supported key types.

### DIFF
--- a/src/identity/client.rs
+++ b/src/identity/client.rs
@@ -7,12 +7,16 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
+use crate::keys::BlsKeypair;
 use crate::{utils, Ed25519Digest, Error, Keypair, PublicKey, Signature, XorName};
+use ed25519_dalek::Keypair as Ed25519Keypair;
 use multibase::Decodable;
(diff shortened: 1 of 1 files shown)


src/identity/client.rs
```diff
@@ -70,6 +74,41 @@ impl FullId {
...
 }
 
+impl From<BlsSecretKey> for FullId {
+    fn from(bls_sk: BlsSecretKey) -> Self {
+        let public = bls_sk.public_key();
+        let keypair = Keypair::Bls(BlsKeypair {
+            secret: SerdeSecret(bls_sk),
+            public,
+        });
+        let public_key = keypair.public_key();
```
mrcnski, line 84: I don't think we need this line, we can just reuse the `public` variable.

mrcnski (approved):
Looks good. Left a minor comment, approving regardless.

src/identity/client.rs
lionel-faber: I believe public is of type `threshold_crypto::PublicKey` and `public_key` is of type `safe_nd::PublicKey`. So we'll need both of them.

PR #136, opened 2019-12-12
chore/ci: add publish test build and caching

mrcnski:
This PR enables caching in GitHub Actions.

I'm happy to report that, while builds are slightly slower if there is no matching cache, they are significantly faster when the cache is available.

[Control build without cache:](https://github.com/m-cat/safe-nd/runs/346206448)

|Ubuntu|Windows|macOS|
|---|---|---|
|145s|206s|123s|

[First build with caching (no cache available):](https://github.com/m-cat/safe-nd/runs/346277957)

|Ubuntu|Windows|macOS|
|---|---|---|
|168s|211s|127s|
|115.8%|102.4%|103.3%|

[Second build with caching (cache available):](https://github.com/m-cat/safe-nd/runs/346522339)

|Ubuntu|Windows|macOS|
|---|---|---|
|55s|92s|48s|
|37.9%|44.6%|39.0%|

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -13,6 +13,8 @@ on:
       - master
 
 env:
+  # Run all cargo commands with --verbose.
+  CARGO_TERM_VERBOSE: true
   RUST_BACKTRACE: 1
 
 jobs:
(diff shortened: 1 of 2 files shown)


S-Coyle (approved):
LGTM

PR #140, opened 2019-12-18
feat/entry: add `version()` method on `SeqEntryAction`

mrcnski:
Adds some useful methods.

Branched off of 0.6.0 because the challenge stuff was breaking SCL.


Yoga07 (approved):
Nice one! LGTM👍

PR #141, opened 2019-12-20
Refactor ImmutableData, step 1.1: Add Blob

oetyng:
Draft for discussions with partial implementations. Not to be merged.

**NB:** The draft is intended to focus on discussion about the **concept changes** and **idiomatic rust**. Small things, like typos, wording in docs, indentation and similar trivialities can be addressed in the real PR (if they are still present). 

For the full implementation, see https://github.com/oetyng/safe-nd/tree/datatypes-refinement-demo

- Adds new module `data`, with `blob.rs`, with same functionality as `immutable_data.rs`.
- Adds the `blob` API side by side with `idata` API.

Step 1.2: Remove `immutable_data.rs` and all references to it, the older naming etc.

### DIFF
--- a/src/data/blob.rs
+++ b/src/data/blob.rs
@@ -199,9 +199,10 @@ impl Kind {
 
     /// Creates `Kind` from a `public` flag.
     pub fn from_flag(public: bool) -> Self {
-        match public {
-            true => Kind::Public,
-            false => Kind::Private,
+        if public {
+            Kind::Public
+        } else {
+            Kind::Private
         }
     }
 }
--- a/src/data/blob.rs
+++ b/src/data/blob.rs
@@ -1,0 +1,419 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+use crate::{utils, Error, PublicKey, XorName};
+use bincode::serialized_size;
+use multibase::Decodable;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::{
+    fmt::{self, Debug, Formatter},
+    u64,
+};
+use tiny_keccak;
+
+/// Maximum allowed size for a serialised blob (ID) to grow to
+pub const MAX_BLOB_SIZE_IN_BYTES: u64 = 1024 * 1024 + 10 * 1024;
+
+/// PrivateBlob: an immutable chunk of data which can be deleted. Can only be fetched
+/// by the listed owner.
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone)]
+pub struct PrivateBlob {
+    /// Address.
+    address: Address,
+    /// Contained data.
+    value: Vec<u8>,
+    /// Contains a set of owners of this data. DataHandlers enforce that a DELETE or OWNED-GET type
+    /// of request is coming from the ClientHandler of the owners.
+    owner: PublicKey,
+}
+
+impl PrivateBlob {
+    /// Creates a new instance of `PrivateBlob`
+    pub fn new(value: Vec<u8>, owner: PublicKey) -> Self {
+        let hash_of_value = tiny_keccak::sha3_256(&value);
+        let serialised_contents = utils::serialise(&(hash_of_value, &owner));
+        let address = Address::Private(XorName(tiny_keccak::sha3_256(&serialised_contents)));
+
+        Self {
+            address,
+            value,
+            owner,
+        }
+    }
+
+    /// Returns the value.
+    pub fn value(&self) -> &Vec<u8> {
+        &self.value
+    }
+
+    /// Returns the set of owners.
+    pub fn is_owner(&self, user: PublicKey) -> bool {
+        self.owner == user
+    }
+
+    /// Returns the set of owners.
+    pub fn owner(&self) -> &PublicKey {
+        &self.owner
+    }
+
+    /// Returns the address.
+    pub fn address(&self) -> &Address {
+        &self.address
+    }
+
+    /// Returns the name.
+    pub fn name(&self) -> &XorName {
+        self.address.name()
+    }
+
+    /// Returns size of contained value.
+    pub fn payload_size(&self) -> usize {
+        self.value.len()
+    }
+
+    /// Returns size of this BlobData after serialisation.
+    pub fn serialised_size(&self) -> u64 {
+        serialized_size(self).unwrap_or(u64::MAX)
+    }
+
+    /// Return true if the size is valid
+    pub fn valid_size(&self) -> bool {
+        self.serialised_size() <= MAX_BLOB_SIZE_IN_BYTES
+    }
+}
+
+impl Serialize for PrivateBlob {
+    fn serialize<S: Serializer>(&self, serialiser: S) -> Result<S::Ok, S::Error> {
+        (&self.value, &self.owner).serialize(serialiser)
+    }
+}
+
+impl<'de> Deserialize<'de> for PrivateBlob {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let (value, owner): (Vec<u8>, PublicKey) = Deserialize::deserialize(deserializer)?;
+        Ok(PrivateBlob::new(value, owner))
+    }
+}
+
+impl Debug for PrivateBlob {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        // TODO: Output owners?
+        write!(formatter, "PrivateBlob {:?}", self.name())
+    }
+}
+
+/// An immutable chunk of data which cannot be deleted.
+#[derive(Hash, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct PublicBlob {
+    /// Network address. Omitted when serialising and calculated from the `value` when
+    /// deserialising.
+    address: Address,
+    /// Contained data.
+    value: Vec<u8>,
+}
+
+impl PublicBlob {
+    /// Creates a new instance of `PublicBlob`
+    pub fn new(value: Vec<u8>) -> Self {
+        Self {
+            address: Address::Public(XorName(tiny_keccak::sha3_256(&value))),
+            value,
+        }
+    }
+
+    /// Returns the value.
+    pub fn value(&self) -> &Vec<u8> {
+        &self.value
+    }
+
+    /// Returns the address.
+    pub fn address(&self) -> &Address {
+        &self.address
+    }
+
+    /// Returns the name.
+    pub fn name(&self) -> &XorName {
+        self.address.name()
+    }
+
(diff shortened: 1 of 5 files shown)


src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+
+    /// Create from z-base-32 encoded string.
+    pub fn decode_from_zbase32<T: Decodable>(encoded: T) -> Result<Self, Error> {
+        utils::decode(encoded)
+    }
+}
+
+/// Object storing an blob variant.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
+pub enum BlobData {
```
mrcnski, line 266: Looks like we did `ImmutableData -> BlobData` instead of `ImmutableData -> Blob`.

oetyng: Great, some reviews coming in!

This is because of `AppendOnlyData`. 
It had `AppendOnly` as the trait, and `AppendOnlyData` as the enum (corresponding to `ImmutableData` / `BlobData` here).

At first, `AppendOnly` trait became `Map` and `Sequence`, which made the distinction necessary in order to compile, calling the enum `MapData` and `SequenceData`, which for consistency required `BlobData` (even though it had no trait `Blob`).

But then I renamed `Map` trait to `MapBase`, and equivalent for `Sequence` (smacks a bit of OO and inheritance, don't really like to call it base or trait - what is idiomatic here?), which would now allow us to ditch "**Data**" on the enums (seems it was overlooked at the time of the rename). 

But I will definitely do that now.

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+use bincode::serialized_size;
+use multibase::Decodable;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::{
+    fmt::{self, Debug, Formatter},
+    u64,
+};
+use tiny_keccak;
+
+/// Maximum allowed size for a serialised blob (ID) to grow to
```
lionel-faber, line 20:
```suggestion
/// Maximum allowed size for a serialised blob (ID) to grow to.
```

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+    pub fn payload_size(&self) -> usize {
+        self.value.len()
+    }
+
+    /// Returns size of this BlobData after serialisation.
+    pub fn serialised_size(&self) -> u64 {
+        serialized_size(self).unwrap_or(u64::MAX)
+    }
+
+    /// Return true if the size is valid
```
lionel-faber, line 85:
```suggestion
    /// Returns true if the size is valid.
```
Just to keep it consistent with the rest of the documentation :)

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+use tiny_keccak;
+
+/// Maximum allowed size for a serialised blob (ID) to grow to
+pub const MAX_BLOB_SIZE_IN_BYTES: u64 = 1024 * 1024 + 10 * 1024;
+
+/// PrivateBlob: an immutable chunk of data which can be deleted. Can only be fetched
+/// by the listed owner.
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone)]
+pub struct PrivateBlob {
+    /// Address.
```
lionel-faber, line 27:
```suggestion
    /// Address. Omitted when serialising and calculated from the `value` and `owner` when
    /// deserialising.
```
I believe that it's nice to have this additonal info. Thoughts?

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+    pub fn value(&self) -> &Vec<u8> {
+        &self.value
+    }
+
+    /// Returns the set of owners.
+    pub fn is_owner(&self, user: PublicKey) -> bool {
+        self.owner == user
+    }
+
+    /// Returns the set of owners.
```
lionel-faber, line 60: I think this can be changed to something like, "Returns the owner's public key"
Or something along those lines, since the functions doesn't exactly return a set.

One more thing we can improve here would be the usage of "owner" vs "owners".
I don't think it's necessary to specify the plural. Since the public key might represent a user / a group of users.

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+    pub fn payload_size(&self) -> usize {
+        self.value.len()
+    }
+
+    /// Returns size of this BlobData after serialisation.
+    pub fn serialised_size(&self) -> u64 {
+        serialized_size(self).unwrap_or(u64::MAX)
+    }
+
+    /// Return true if the size is valid
```
lionel-faber, line 155:
```suggestion
    /// Returns true if the size is valid.
```

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+    pub fn is_private(&self) -> bool {
+        self.kind().is_private()
+    }
+
+    /// Returns the Address serialised and encoded in z-base-32.
+    pub fn encode_to_zbase32(&self) -> String {
+        utils::encode(&self)
+    }
+
+    /// Create from z-base-32 encoded string.
```
lionel-faber, line 258:
```suggestion
    /// Creates from z-base-32 encoded string.
```

lionel-faber (changes requested):
Nice work @oetyng!
I've left a few comments as well.
I don't think we'll need Blob and IData side by side. IData can be removed so that the changes in SCL and vault will be just renaming.

src/data/blob.rs
oetyng: Yep, this is part of docs/code that has not been modified in this work.
Maybe hard to see when there has been renaming all over.

src/data/blob.rs
oetyng: Yep sure, this is part of docs/code that has not been modified in this work.
Maybe hard to see when there has been renaming all over.

src/data/blob.rs
oetyng: Yep, agree. This one is a remnant from the old code. I have been changing from plural to singular in all code I've changed (afaik).

src/data/blob.rs
oetyng: Absolutely, I see that's what it says on PublicBlob. I think all comments in this file have been copied straight over. The name changes have been done with find and replace.

lionel-faber:
[Here's a diff](https://www.diffchecker.com/gzBNhHGG) that I found quite useful when reviewing :)

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+            owner,
+        }
+    }
+
+    /// Returns the value.
+    pub fn value(&self) -> &Vec<u8> {
+        &self.value
+    }
+
+    /// Returns the set of owners.
```
mrcnski, line 55: This doc comment should be udpated.

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+        self.value.len()
+    }
+
+    /// Returns size of this BlobData after serialisation.
+    pub fn serialised_size(&self) -> u64 {
+        serialized_size(self).unwrap_or(u64::MAX)
+    }
+
+    /// Return true if the size is valid
+    pub fn valid_size(&self) -> bool {
```
mrcnski, line 156: It's standard practice for boolean functions to start with a question word and it is a bit more clear when using such functions. So here I would name the function something like `has_valid_size`.

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+    }
+}
+
+impl Debug for PublicBlob {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(formatter, "PublicBlob {:?}", self.name())
+    }
+}
+
+/// The Kind determines if a blob is public or private.
```
mrcnski, line 180: I'm not sure if "determines" is the best word here

mrcnski (changes requested):

src/data/blob.rs
mrcnski: I like to keep verbs in the third person to be consistent with the Rust standard library documentation

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
+// Copyright 2019 MaidSafe.net limited.
```
bochaco, line 1: it seems al will need to be 2020 ?

src/data/blob.rs
```diff
@@ -0,0 +1,420 @@
...
+        self.address.name()
+    }
+
+    /// Returns size of contained value.
+    pub fn payload_size(&self) -> usize {
+        self.value.len()
+    }
+
+    /// Returns size of this BlobData after serialisation.
+    pub fn serialised_size(&self) -> u64 {
```
bochaco, line 81: It sounds like we need a Result as return type rather than returning a u64::MAX

src/data/blob.rs
bochaco: How about using/reserving `Blob` only for the old `ImmutableData` which contains the datamap of chunks with self-encryption algo, and here we then have `PrivateChunk`/`PublicChunk` instead of the `PrivateBlob`/`PublicBlob`, and `Chunk` instead of `BlobData` ?
So `Blob` is only the user/API facing concept, then we have `Chunks`

src/data/blob.rs
oetyng: This method looks weird. But it is old code. I haven't changed anything but naming in this file.

src/data/blob.rs
oetyng: @bochaco , if I understood you correctly, that would introduce a different pattern than what we do for Map and Sequence.
Have a look at these enums in safe-nd PR #145.

bochaco: Yes, but we could align them also to this other pattern, specially if we consider the other proposal about DataMaps and Chunks, a Blob/Map/Sequence is what we expose as the data types, then each of them is composed of different structures linking chunks. We are otherwise kinda overriding the meaning of a "Blob", is a blob a chunk but also in some other parts a group of chunks with a self-encryption algo to compose what it used to be an ImmutableData file? this is where i see the issue with reusing the name blob for the priv/pub chunk and the actual Blob composed of chunks

oetyng: Oh wow, I had actually misunderstood this data type. It's not the Blob type there, it's just a chunk.. I did not intend for a chunk to be named Blob. 

Yes, this definitely has to be different.

Well, with the other proposal, this also changes. There are no Private or Public chunks, just chunks. Hm, something needs to be done here, but can't really adapt it to the other proposal now.

bochaco: yeah, but you could even consider a priv/pub chunk enum type to be with/without ref-count respectively rather than having the ref-count to be Option<...>

oetyng: Mm, there's a slight break in logic without considering DataMap/Chunk proposal.
But, it's ok. 

From `lib.rs`:

```
pub enum DataType {
    Blob(Chunk),
    Map(Map),
    Sequence(Sequence),
}
```

And it'll be `PrivateChunk`/`PublicChunk`. But I'll do that in the real PR (#147).

src/data/blob.rs
oetyng: I am aware of that. I have never used any other style (the rest of the data types consistently use **is_**, **has_** etc.) so was trying to remember why this got it. The only thing I can come think of is that this is the very first part I started (Blob), and I saw short forms of things I normally see prefixed with intention - like **get**; so, `value()` instead of `get_value()`, or` permissions(`) instead of `get_permissions()`. So I think that came as I perceived a code style there that I just didn't want to change too much.

As I got into the other types, there was already so much fixing that such a change seemed small, and I probably never looked back at that one.

PR #142, opened 2019-12-20
Refactor Permissions, Step 2.1: Add access_control

oetyng:
Draft for discussions with partial implementations. **Not to be merged.**

**NB:** The draft is primarily intended to focus on discussion about the **concept changes** and **idiomatic rust**. Small things, like typos or smallish wording changes in docs, indentation and similar trivialities can be addressed in the real PR (if they are still present). 

For the full implementation, see https://github.com/oetyng/safe-nd/tree/datatypes-refinement-demo

- Adds `shared_types.rs`.
- Adds `access_control.rs` in `authorization` mod.
- Identifies existing permissions concept as access list.

Step **2.2** adds tests using the refactored data types.

### DIFF
--- a/src/authorization/access_control.rs
+++ b/src/authorization/access_control.rs
@@ -1,0 +1,209 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+use crate::shared_types::User;
+use crate::PublicKey;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::{collections::BTreeMap, hash::Hash};
+
+/// ===========================================================
+///  Access control of data type instances and their content.
+/// ===========================================================
+
+/// The type of access to the native data structures.
+#[derive(Clone, Copy, Debug, Hash, Eq, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub enum AccessType {
+    /// Read data, owners and permissions.
+    Read,
+    /// Append new values.
+    Append,
+    /// Insert new values.
+    Insert,
+    /// Soft-update existing values.
+    Update,
+    /// Soft-delete existing values.
+    Delete,
+    /// Hard-update existing values.
+    HardUpdate,
+    /// Hard-delete existing values.
+    HardDelete,
+    /// Modify permissions for other users.
+    ModifyPermissions,
+}
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash, Debug)]
+pub enum AccessList {
+    Public(PublicAccessList),
+    Private(PrivateAccessList),
+}
+
+impl From<PrivateAccessList> for AccessList {
+    fn from(list: PrivateAccessList) -> Self {
+        AccessList::Private(list)
+    }
+}
+
+impl From<PublicAccessList> for AccessList {
+    fn from(list: PublicAccessList) -> Self {
+        AccessList::Public(list)
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash, Debug)]
+pub enum UserAccess {
+    Public(PublicUserAccess),
+    Private(PrivateUserAccess),
+}
+
+impl From<PrivateUserAccess> for UserAccess {
+    fn from(access: PrivateUserAccess) -> Self {
+        UserAccess::Private(access)
+    }
+}
+
+impl From<PublicUserAccess> for UserAccess {
+    fn from(access: PublicUserAccess) -> Self {
+        UserAccess::Public(access)
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash, Debug)]
+pub struct PrivateUserAccess {
+    status: BTreeMap<AccessType, bool>,
+}
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash, Debug)]
+pub struct PublicUserAccess {
+    status: BTreeMap<AccessType, bool>,
+}
+
+impl PrivateUserAccess {
+    pub fn new(status: BTreeMap<AccessType, bool>) -> Self {
+        PrivateUserAccess { status }
+    }
+
+    pub fn set(&mut self, status: BTreeMap<AccessType, bool>) {
+        self.status = status;
+    }
+
+    pub fn is_allowed(self, access: &AccessType) -> bool {
+        match self.status.get(access) {
+            Some(true) => true,
+            _ => false,
+        }
+    }
+}
+
+impl PublicUserAccess {
+    pub fn new(status: BTreeMap<AccessType, bool>) -> Self {
+        PublicUserAccess { status }
+    }
+
+    pub fn set(&mut self, status: BTreeMap<AccessType, bool>) {
+        self.status = status; // todo: filter out Queries
+    }
+
+    /// Returns `Some(true)` if `access` is allowed and `Some(false)` if it's not.
+    /// `None` means that `User::Anyone` permissions apply.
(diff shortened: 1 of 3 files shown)


src/shared_types.rs
```diff
@@ -0,0 +1,213 @@
...
+// Software.
+
+use crate::{utils, PublicKey, Result, XorName};
+use multibase::Decodable;
+use serde::{Deserialize, Serialize};
+use std::ops::Range;
+
+pub type Key = Vec<u8>;
+pub type Value = Vec<u8>;
+pub type KvPair = (Key, Value);
```
mrcnski, line 17: Is this type being used anywhere? Also, I think `KeyValue` might be a better name.

src/shared_types.rs
```diff
@@ -0,0 +1,213 @@
+// Copyright 2019 MaidSafe.net limited.
```
mrcnski, line 1: I like the idea of a shared types module!

src/shared_types.rs
```diff
@@ -0,0 +1,213 @@
...
+use serde::{Deserialize, Serialize};
+use std::ops::Range;
+
+pub type Key = Vec<u8>;
+pub type Value = Vec<u8>;
+pub type KvPair = (Key, Value);
+pub type Values = Vec<Value>;
+pub type Keys = Vec<Key>;
+
+/// Marker for sentried data.
```
mrcnski, line 21: Might be worth documenting here again what "sentried" means.

src/shared_types.rs
```diff
@@ -0,0 +1,213 @@
...
+        self.expected_data_version
+    }
+
+    pub fn expected_owners_version(&self) -> u64 {
+        self.expected_owners_version
+    }
+
+    pub fn expected_access_list_version(&self) -> u64 {
+        self.expected_access_list_version
+    }
```
mrcnski, line 78: I would just make the struct fields public instead of having these getters -- unless you don't want the fields being modified?

src/shared_types.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, Debug)]
+pub struct Owner {
+    pub public_key: PublicKey,
+    /// The expected Version of the data at the time this ownership change is to become valid.
+    pub expected_data_version: u64,
```
mrcnski, line 85: I'm wondering whether we can define `type Version = u64;` and use `Version` everywhere. We just need to pick a different name since `Version` is taken.

src/shared_types.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// The expected Version of the permissions at the time this ownership change is to become valid.
+    pub expected_access_list_version: u64,
+}
+
+#[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
+pub enum Kind {
```
mrcnski, line 91: This doesn't really fit for immutable data -- sorry, blobs -- which only has two kinds. If we're not planning to use it for blobs, then maybe rename this `SentriedKind`?

src/authorization/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+pub mod access_control;
+mod tests;
```
mrcnski, line 11: We like to keep separation between `pub mod` and `mod` statements because one exports an API publicly and one doesn't. It's useful to keep them in separate blocks so it's easier to see the public exports at a glance.

src/authorization/access_control.rs
```diff
@@ -0,0 +1,209 @@
...
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::{collections::BTreeMap, hash::Hash};
+
+/// ===========================================================
+///  Access control of data type instances and their content.
+/// ===========================================================
+
+/// The type of access to the native data structures.
+#[derive(Clone, Copy, Debug, Hash, Eq, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub enum AccessType {
```
mrcnski, line 21: Can we name this `AccessKind` or even just `Kind` (and export it as `AccessKind`) to be more consistent with using `Kind` elsewhere?

src/authorization/access_control.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Read data, owners and permissions.
+    Read,
+    /// Append new values.
+    Append,
+    /// Insert new values.
+    Insert,
+    /// Soft-update existing values.
```
mrcnski, line 28: Would like to eventually see some documentation regarding the difference between soft- and hard-updates

src/authorization/access_control.rs
```diff
@@ -0,0 +1,209 @@
...
+pub struct PublicUserAccess {
+    status: BTreeMap<AccessType, bool>,
+}
+
+impl PrivateUserAccess {
+    pub fn new(status: BTreeMap<AccessType, bool>) -> Self {
+        PrivateUserAccess { status }
+    }
+
+    pub fn set(&mut self, status: BTreeMap<AccessType, bool>) {
```
mrcnski, line 91: I don't really see the point in having a `set` method when someone could just as easily create a new `UserAccess` instance. Also, making the `status` field public would seem to obviate the need for this as well as the `new` constructor.

mrcnski (changes requested):
Left some comments with my thoughts 🙂 Can you link from this PR to a document describing how the new access controls work? Also, eventually making the in-code documentation more robust would be great 👍

src/shared_types.rs
oetyng: Yep, it is used. 
About the name, yeah `Kv` is not so nice. I did use `KeyValue` here at some point as well and I generally don't like abbreviations. But anyway, I think `KeyValuePair` is most expressive, since `KeyValue` could be meaning `the value of a key` and isn't necessarily `the key and its value`. So, less unambiguous with `Pair` in there IMO.

src/authorization/mod.rs
oetyng: You mean, in this case separation with a line?

src/shared_types.rs
mrcnski: Agreed, I like `KeyValuePair`!

src/authorization/mod.rs
mrcnski: Exactly! [Example](https://github.com/maidsafe/safe-nd/blob/master/src/lib.rs)

src/authorization/access_control.rs
oetyng: This is how it looked in Permissions already (different name though). But yeah, I agree, that's a bit strange actually, not so functional. 

The reason I think was that in the `PublicUserAccess`, there is logic for filtering out queries (originally, I did not implement that filter here yet).
But, that logic could just as well go in the ctor. 

> making the `status` field public would seem to obviate the need for this as well as the `new` constructor.
Public field wouldn't be desired here I'd say.

I think moving the query filter to ctor and always instantiate new instead of modifying an instance would be best here?

src/shared_types.rs
oetyng: I'm not sure what the intention was to begin with. This is also just renamed code.
But personally, I would consider most structs like these to be immutable [_value objects_](https://deviq.com/value-object/) (as in DDD).
So no modification of the fields.

Normally, I would do the fields `public` and `readonly` in C#, but that doesn't work in Rust. I do not like getters like that, it's super bloaty (very Java-ish as well :) ).

src/authorization/access_control.rs
mrcnski: Ah, that makes sense now. I didn't know there was going to be filtering logic.

> I think moving the query filter to ctor and always instantiate new instead of modifying an instance would be best here?

I agree!

src/shared_types.rs
oetyng: Yeah the `Address` and `Kind` thing being different for `Blob` was something I didn't quite like as well.
It's one of those small things in the path towards a more stringent data type handling that I undertook (the symlink/metadata idea).

But yeah, `SentriedKind` would be a better distinction as long as there is a difference.

src/shared_types.rs
mrcnski: Okay, that makes sense.

src/authorization/access_control.rs
oetyng: Yeah, that's better.

PR #144, opened 2019-12-20
Refactor AppendOnlyData, step 4.1: Add Map

oetyng:
Draft for discussions with partial implementations. **Not to be merged.**

**NB:** The draft is primarily intended to focus on discussion about the **concept changes** and **idiomatic rust**. Small things, like typos or smallish wording changes in docs, indentation and similar trivialities can be addressed in the real PR (if they are still present). 
Anything off that is already present in current code base (i.e. not part of this change) could perhaps better go into a proper issue - as to keep discussion focused around the actual changes.

For the full implementation, see https://github.com/oetyng/safe-nd/tree/datatypes-refinement-demo

Based on `access_control` branch.

- Adds `map.rs` in `data` mod.
- Adds rudimentary tests for `map`.

Step **4.2** adds the `request` and `response` APIs for the `map` data type.

PR #87, opened 2019-07-29
Simplify adata and mdata implementations

madadam:
The prevous impl based on traits and macros was unnecessarily complex. This PR simplifies it using an approach similar to "policy-based design" known from C++.

PR #145, opened 2019-12-23
Datatypes refinement demo (reference only)

oetyng:
A demo of the full data types refinement change, for reference when reviewing the more granular draft PRs:

- https://github.com/maidsafe/safe-nd/pull/141
- https://github.com/maidsafe/safe-nd/pull/142
- https://github.com/maidsafe/safe-nd/pull/143
- https://github.com/maidsafe/safe-nd/pull/144

**Not to be merged.**

**NB:** The draft is primarily intended to focus on discussion about the **concept changes** and **idiomatic rust**. Small things, like typos or smallish wording changes in docs, indentation and similar trivialities can be addressed in the real PR (if they are still present). 


src/authorization/access_control.rs
```diff
@@ -0,0 +1,209 @@
...
+pub struct PublicAccessList {
+    pub access_list: BTreeMap<User, PublicUserAccess>,
+    /// The expected index of the data at the time this grant status change is to become valid.
+    pub expected_data_version: u64,
+    /// The expected index of the owners at the time this grant status change is to become valid.
+    pub expected_owners_version: u64,
+}
+
+impl PublicAccessList {
+    fn is_allowed_(&self, user: &User, access: AccessType) -> Option<bool> {
```
maqi: not sure, however this `is_allowed_` nested whitin `is_allowed` doesn't seem a good practice?

src/data/blob.rs
```diff
@@ -0,0 +1,337 @@
...
+/// by the listed owner.
+#[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone)]
+pub struct PrivateBlob {
+    /// Address.
+    address: Address,
+    /// Contained data.
+    value: Vec<u8>,
+    /// Contains a set of owners of this data. DataHandlers enforce that a DELETE or OWNED-GET type
+    /// of request is coming from the ClientHandler of the owners.
+    owner: PublicKey,
```
maqi, line 34: shall use a vector?

src/data/blob.rs
```diff
@@ -0,0 +1,337 @@
...
+impl Debug for PrivateBlob {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        // TODO: Output owners?
+        write!(formatter, "PrivateBlob {:?}", self.name())
+    }
+}
+
+/// An immutable chunk of data which cannot be deleted.
+#[derive(Hash, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct PublicBlob {
```
maqi: maybe better have `PrivateBolb` & `PublicBolb` as:
```rust
struct Bolb {
    address: Address,
    value: Vec<u8>,
}

struct PrivateBolb {
    bolb: Bolb,
    owners: Vec<PublicKey>,
}

struct PublicBolb {
    bolb: Bolb,
}
```

src/shared_data.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+}
+
+// pub type Key = Vec<u8>;
+pub type Value = Vec<u8>;
+pub type KvPair = (Key, Value);
+pub type Values = Vec<Value>;
+pub type Keys = Vec<Key>;
+
+/// Marker for sentried data.
```
maqi: any explanation of what is `sentried` and `non-sentried` data?

src/shared_data.rs
```diff
@@ -0,0 +1,237 @@
...
+
+impl From<u64> for Version {
+    fn from(version: u64) -> Self {
+        Version::FromStart(version)
+    }
+}
+
+// Set of data, owners, permissions versions.
+#[derive(Copy, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+pub struct ExpectedVersions {
```
maqi: why called `ExpectedVersions` ? this seems not much generic ?

src/shared_data.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub fn expected_access_list_version(&self) -> u64 {
+        self.expected_access_list_version
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, Debug)]
+pub struct Owner {
+    pub public_key: PublicKey,
+    /// The expected Version of the data at the time this ownership change is to become valid.
+    pub expected_data_version: u64,
```
maqi: this means the data is owned by an owner only if the version of the data is greater than the expected_data_version? or less than it?
wondering what if a data's previous owner got removed and then added back later on?

just not clear about the usage of this `expected_data_version` and the below `expected_access_list_version`

src/shared_data.rs
```diff
@@ -0,0 +1,237 @@
...
+}
+
+#[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
+pub enum DataAddress {
+    Map(Address),
+    Sequence(Address),
+}
+
+#[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
+pub enum Address {
```
maqi: why we need to differentiate them?

src/response.rs
```diff
@@ -8,51 +8,74 @@
...
+    GetMapValueAt(Result<Value>), // The value of a key as of a version.
+    GetMapValues(Result<Values>), // All current values of map
+    GetMapEntries(Result<MapEntries>),
+    GetMapKeys(Result<Keys>),
+    GetMapKeyHistories(Result<MapKeyHistories>),
+    GetMapKeyHistory(Result<MapValues>),
+    GetMapKeyHistoryRange(Result<MapValues>),
+    GetMapOwner(Result<Owner>),
+    GetMapOwnerAt(Result<Owner>),
+    GetMapOwnerHistory(Result<Vec<Owner>>),
```
maqi, line 44: this method sounds like trying to fetch the history of the owner, i.e. includes the owner that got removed.
or, it means to return the `owners list` ?

src/data/sequence.rs
```diff
@@ -0,0 +1,788 @@
...
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash)]
+pub struct SequenceBase<P, S> {
+    address: Address,
+    data: Values,
+    access_list: Vec<P>,
+    // This is the history of owners, with each entry representing an owner.  Each single owner
+    // could represent an individual user, or a group of users, depending on the `PublicKey` type.
+    owners: Vec<Owner>,
+    _flavour: S,
```
maqi: not used? then why we need this extra type parameter?

src/data/sequence.rs
```diff
@@ -0,0 +1,788 @@
...
+
+/// Common methods for all `Sequence` flavours.
+impl<C, S> SequenceBase<C, S>
+where
+    C: AccessListTrait,
+    S: Copy,
+{
+    /// Returns true if the provided access type is allowed for the specific user (identified y their public key).
+    pub fn is_allowed(&self, user: PublicKey, access: AccessType) -> bool {
+        if let Some(owner) = self.owner_at(Version::FromEnd(1)) {
```
maqi: this manual version value seems suspicious. 
at least a comment explaining why we can do so?

src/data/sequence.rs
```diff
@@ -0,0 +1,788 @@
...
+    }
+
+    /// Return the type tag of this Sequence.
+    pub fn tag(&self) -> u64 {
+        self.address.tag()
+    }
+
+    /// Returns true if the user is the current owner.
+    pub fn is_owner(&self, user: PublicKey) -> bool {
+        match self.owner_at(Version::FromEnd(1)) {
```
maqi: so only one owner allowed?

src/data/map.rs
```diff
@@ -0,0 +1,1561 @@
...
+                };
+                (insert, update, delete)
+            },
+        );
+
+        self.hard_apply(operations)
+    }
+
+    fn hard_apply(&mut self, operations: SentriedOperations) -> Result<()> {
+        let (insert, update, delete) = operations;
```
maqi: this usage is bit risky, especially `insert` and `update` has the same type.
can we make `SentriedOperations` as a proper struct and having a function returns insert/update/delete ?

src/data/map.rs
```diff
@@ -0,0 +1,1561 @@
...
+                        let _ = insert.insert(kv_pair.clone());
+                    }
+                    Cmd::Update(kv_pair) => {
+                        let _ = update.insert(kv_pair.clone());
+                    }
+                    Cmd::Delete(key) => {
+                        let _ = delete.insert(key.clone());
+                    }
+                };
+                (insert, update, delete)
```
maqi: will this lose the sequence of interleaving?
for example `insert A, delete A, insert A, update A` will be applied as: insert A, then update A, then delete A ?

src/data/sequence.rs
oetyng: BLS key gives multiple owner under 1 key. This is just copied from previous implementation, nothing changed, except from wording. It was referred to as "owners" before, and `append_owner`, which would give impression of adding to existing owners, but validation of ownership was always made against _current entry_. The vector of owner was in fact the _history_ of owners, but it was not expressed in the naming of operations (therefore I renamed `append_owner` to `set_owner`). So, it has always been one owner-key always. Only with BLS shares can it currently be multiple owners. This needs clarification of how to be handled end to end, or if we want to do something differently.

src/data/map.rs
oetyng: Yes well spotted, this is a problem discussed in the pre-rfc here: https://github.com/maidsafe/pre-rfc/pull/112#discussion_r351410563
I suggest there that it should be fixed.

src/data/sequence.rs
oetyng: Yeah I know and agree. (Additionally, it might actually be FromEnd(0) to get last entry of the vec..)
The aim is top get the `current entry`, which is the current owner, without specifying an index.
It would only work if there are no race conditions. Do you have suggestion for this?

src/shared_data.rs
oetyng: It's per this design change: https://github.com/maidsafe/safe-nd/pull/126 (contains rationale)
... where recently (Sunday December 8th, in #safe-core) I discussed with @lionel1704 to change that to `Versions` instead of `Indices`:

> @lionel Regarding, the expected indices PR:
> 
> After having worked with the data types quite intensely now for a while, I've learned more about the usage of version and index at various places.
> I can see that it is used interchangeably for the same thing, where version is the concept which we IMO directly associate with the functionality of concurrency control and data perpetuity, and index is the implementation detail which reveals that we're working against a vector of sorts.
> I am quite strongly suggesting that we go consistent here and use only one of these all over, and also that it be version that we use (which was the case previously with MutableData), and keep the implementation detail from leaking through the layers.
> 
> This would in practice mean that the Index data struct is simply renamed to Version. And henceforth we talk about current version and expected version.
> I am absolutely positive that this corresponds better to the domain which it is used in (concurrency control & data perpetuity).
> Internally in safe-nd we can still map it to an Index type if we want to have a gradual transition to a language which best corresponds to that level of abstraction - i.e. when we're in the implementation details of the data types, we do know that we're dealing with indices.

where he replied

> You're absolutely right Edward. Index and version are used interchangeably. They can be renamed to version alone. Makes a lot of sense.

----------------------------------

When you say not much generic, could you please elaborate on what change you'd like to see and why?

src/shared_data.rs
oetyng: It's all described in the DataTypes Refinement Pre-RFC, there's both the HackMD and the PR to pre-rfc repo, I can link if you want?

src/data/blob.rs
oetyng: You mean for PublicKey? I don't know, nothing changed there from previous implementation.

src/data/map.rs
oetyng: Yes well spotted, this is a problem discussed in the pre-rfc.
I suggest there that it should be fixed:

> We would need to consider what behaviour we want to support, and adapt network side handling to it. If we want to be able to do this:
> 
> ```
> let mymap = new Map();
> mymap.startTx();
> mymap.insert("key1", "value1");
> mymap.delete("key1");
> mymap.insert("key1", "value2"); // <---- insert same key again
> mymap.commit();
> ```
> 
> or even
> 
> ```
> let mymap = new Map();
> mymap.startTx();
> mymap.insert("key1", "value1");
> mymap.delete("key1");
> mymap.insert("key1", "value1"); // <--  insert same key again, with same value.
> mymap.commit();
> ```
> 
> .. then we have to change some of the current logics.
> Since it applies all inserts in a group, then all updates and then all deletes, the first example would blow up since it tries to perform
> ```
> mymap.insert("key1", "value1");
> mymap.insert("key1", "value2");
> mymap.delete("key1");
> ```
> 
> But, that's not a big thing to fix, we would then group operations by key, and _then_ apply them in the order they were added to the tx.
> 
> The second example would not result in "key1" = "value1", even if we fixed the key-grouping as per above, since the last operation - with current logics - would be filtered out, so we would end up with the key being deleted. Also not a big thing to fix, but we just need to be aware of things like this :)
>

src/data/blob.rs
maqi: Yes, for PublicKey.
well, the comment says `Contains a set of owners of this data`
maybe this shall be BlsKey ?

src/data/sequence.rs
oetyng: Good question. I would think it is to be able to differentiate the trait, and do different implementations based on the flavour.
Not sure why it would be stored in a field though, perhaps it has to for compiler to agree? I reused this solution from @madadam 's PR: https://github.com/maidsafe/safe-nd/pull/87

src/authorization/access_control.rs
oetyng: Mm I agree it's an iffy naming practice. Could need some change, I don't remember quite now why it was done like this. Will look it up.

src/shared_data.rs
maqi: Ah, when I say `not much generic`, I just mean the name is `Expected`Version, meanwhile the content it holds is actually just version info, and could be used for not only `expected` purpose. (could be current, or last, ...)

src/data/map.rs
oetyng: Yeah, that would seem better, I'll try something out for it.

src/data/sequence.rs
maqi: I don't have a better solution to the issue, however, I'd at least use a const and put some comment to explain why doing so.
so that, in case we do need to change from `FromEnd(1)` to `FromEnd(0)`, it will be more explicit.

src/data/blob.rs
oetyng: Mm, perhaps, but it would break the pattern used for `Map` and `Sequence` (in code currently called `MapData` and `SequenceData`, a remnant of a previous a naming conflict, but "**Data**" is to be scrapped since the naming conflict was solved).

```
/// Object storing a Map variant.
pub enum MapData {
    /// Public instance with concurrency control.
    PublicSentried(PublicSentriedMap),
    /// Public instance.
    Public(PublicMap),
    /// Private instance with concurrency control.
    PrivateSentried(PrivateSentriedMap),
    /// Private instance.
    Private(PrivateMap),
}
```

I'd prefer keeping the pattern between the data types for expressing the same kind of phenomenon.

src/data/sequence.rs
maqi: but the type of user is `PublicKey`, which normally means `a public key for just one user`
If we do want to use BLS key (or plan to do so later on), maybe change the type to something like `OwnershipKey`, and currently define `type OwnershipKey = PublicKey`, and later on switch it to BLSKey ?

oetyng: Yep, I think that's a good idea.

src/shared_data.rs
oetyng: I'm not entirely sure of this myself. It's also something kept from previous implementation.
I think the end result is that these get different name spaces, so you could store data under same xorname+tag for all four kinds. I'm not totally clear what benefits and drawbacks we have with that.
 Perhaps motivation is found in the RFCs for the Published/Unpublished, Sequenced/Unsequenced data types, otherwise, I think @nbaksalyar and others might have the exact background.

src/response.rs
oetyng: Ah yes, I had planned to fix this, `owners` field in `MapBase` should be called `owner_history`, and `access_list` in the same way `access_list_history`.

These are append only, so the act of removing an owner is to set a new owner.

Naming the list `owners` is a remnant of previous implementation, which gave the impression that they were all currently owners. But validation of ownership only ever looked at the last entry in the vector, so it is indeed a history of owners.

src/shared_data.rs
oetyng: If I understand your question correctly: no.

Except for naming, the implementation is the same now as before. I have also thought slightly about if something should change here, but it ended up fairly low on my priority list.

The Owner struct is used both as a DTO and as a domain model, which is a bit iffy IMO, because it is used for slightly different reasons, which are not apparent (as evidenced by your question as well).

As a DTO in a request to set the owner, the `expected_data_version` and `expected_access_list_version` acts as optimistic concurrency control, i.e. if anything changed in the data or permissions from the point that the requester sees it, to when the request is handled, it is rejected.
As a domain model, it tells the history of who the owner was at a particular data version. This would presumably let us answer questions like: `when current data version was x, who was the owner? when this set of users had this set of permissions (i.e. access list version y), who was the owner?`

src/data/blob.rs
oetyng: I think the end goal was this, to have it as `BlsKey`. I did ask around in slack a few weeks back for this, unfortunately I couldn't get it cleared out at the time. 
But I see no apparent reason why not doing it, I have always assumed `BlsKeys` will be the way we have multiple ownership of any and everything in the network.

src/shared_data.rs
oetyng: Ah OK I see. Mm that's true. I think before it was often even just an `u64`, so at least it's a bit more readable now. But I guess static type checking on the actual purpose could be implemented as well.
Do you propose an `ExpectedVersion` type, that can only be used for that purpose?

maqi: > Do you propose an ExpectedVersion type, that can only be used for that purpose?

Yes, I'd prefer to do so.

src/data/sequence.rs
oetyng: Saw now that the question of PublicKey is moot, since it already is defined like this:

```
pub enum PublicKey {
    Ed25519(ed25519_dalek::PublicKey),
    Bls(threshold_crypto::PublicKey),
    BlsShare(threshold_crypto::PublicKeyShare),
}
```

src/data/blob.rs
oetyng: Saw now that the question of PublicKey is moot, since it already is defined like this:

```
pub enum PublicKey {
    Ed25519(ed25519_dalek::PublicKey),
    Bls(threshold_crypto::PublicKey),
    BlsShare(threshold_crypto::PublicKeyShare),
}
```

But docs could be improved as you say.

maqi: Ah, that will be great.

src/data/sequence.rs
maqi: I am wondering that whether the network can allow different type of `PublicKey` co-exists within the network. (i.e. some node using `Ed25519` as PublicKey, meanwhile others using `Bls`).
Also, not sure about the diff between `Bls` and `BlsShare`, can they co-exist?
So, maybe making `PublicKey` as enum is un-necessary?

bochaco: I'd also like to get a clarification about this, or do we need to use specific type of keys in different areas of the network, e.g. a SafeKey (CoinBalance) always is on a BLS PK, so all code related to managing this type of data uses a PublicKey::Bls/BlsShare, etc.

Moreover, why is it we need the Ed25519? can't we use BLS for everything?

src/authorization/access_control.rs
oetyng: The problem is the lack of method overloading in rust.
The one with underscore is a private one that is called for a more specific case.
I tried putting all validation in the trait impl, but that looked very messy.
I think that underscore is the lesser bad of the two.
But some distinguishing word could be added of course.

jeanphilippeD: Yes, adding a word expressing the more specific case would be good. I don't seem to have seen such naming otherwise in rust.

src/authorization/access_control.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    pub fn access_list(&self) -> &BTreeMap<User, PublicUserAccess> {
+        &self.access_list
+    }
+}
+
+impl AccessListTrait for PublicAccessList {
+    fn is_allowed(&self, user: &PublicKey, access: AccessType) -> bool {
+        match self.is_allowed_(&User::Specific(*user), access) {
```
jeanphilippeD: Probably more idomatic and clearer:

```rust
     fn is_allowed(&self, user: &PublicKey, access: AccessType) -> bool {
        self.is_allowed_(&User::Specific(*user), access)
            .or_else(|| self.is_allowed_(&User::Anyone, access))
            .unwrap_or(false)
    } 
```

src/authorization/access_control.rs
oetyng: I was actually wrong, the private one is the less specific.
The hard part though is that there is no difference in semantics so the extra word will be very artificial.

I don't know how you usually do, but adding the types of the parameters in the function name is not a very elegant style. The extension of that is to not name things for what they represent, i.e. intent, (user for PublicKey) but always for what they are implemented with (implementation detail).

The public one checks if the user (represented as public key) access combo is allowed.
The private one checks if the user (represented by enum User) access combo is allowed.

Feel free to drop suggestions here.

src/authorization/access_control.rs
oetyng: This is exactly the kind of comments I was hoping for. Great!

src/authorization/access_control.rs
jeanphilippeD: If there is really no need for having different name (not 100% convinced, is_allowed is so very generic seem we could be a bit more specific, but also the function seem really simple so maybe it is fine (Yes, adding type parameter is not a great idea in general)).

I think you need not use `_` though:
https://doc.rust-lang.org/book/ch19-03-advanced-traits.html

> Fully Qualified Syntax for Disambiguation: Calling Methods with the Same Name
Nothing in Rust prevents a trait from having a method with the same name as another trait’s method, nor does Rust prevent you from implementing both traits on one type. It’s also possible to implement a method directly on the type with the same name as methods from traits.

src/authorization/access_control.rs
```diff
@@ -0,0 +1,209 @@
...
+impl PrivateAccessList {
+    pub fn access_list(&self) -> &BTreeMap<PublicKey, PrivateUserAccess> {
+        &self.access_list
+    }
+}
+
+impl AccessListTrait for PrivateAccessList {
+    fn is_allowed(&self, user: &PublicKey, access: AccessType) -> bool {
+        match self.access_list.get(user) {
+            Some(access_status) => access_status.clone().is_allowed(access),
```
jeanphilippeD: So like the other comment, a lot of time when reaching for a `match` it is better to look for chaining functions.
I'll let you look for the other cases. the option doc is pretty good here https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or

can think a bit of option as an iterator of one or zero element and treat them the same (can even convert option into an iterator if needed).

```rust
     fn is_allowed(&self, user: &PublicKey, access: AccessType) -> bool {
        // is the clone really needed?
        self.access_list.get(user)
            .map(|access_status| access_status.clone().is_allowed(access))
            .unwrap_or(false)
    }
```

src/data/map.rs
```diff
@@ -0,0 +1,1593 @@
...
+    /// Returns true if this instance employs concurrency control.
+    pub fn is_sentried(&self) -> bool {
+        self.kind().is_sentried()
+    }
+
+    /// Returns true if the provided user (identified by their public key) is the current owner.
+    pub fn is_owner(&self, user: PublicKey) -> bool {
+        use Map::*;
+        match self {
+            PublicSentried(data) => data.is_owner(user),
```
jeanphilippeD: considering it is repeated again and again, might be worth using a macro like https://github.com/maidsafe/routing/blob/fleming/src/state_machine.rs#L36 (but a bit simpler), it will make all these function trivial and avoid duplication.

src/authorization/access_control.rs
oetyng: Aah, bad me..! I didn't check the logics or run the tests before committing this change :). 

It actually does not do the same thing.

src/authorization/access_control.rs
oetyng: Yep, I saw this as well. Just seemed like a lot of bloat with a couple of traits + impl, when an underscore solves it. But, it is more correct of course. 
Not sure here, because I'm not a fan of that underscore, but sometimes I think terse and slim wins over "correct".

oetyng:
This is (roughly) superseded by PR #163 (and expected PRs for coming CRDT data types).
A new take on the peripheral changes included in this PR needs to be done after CRDT data types are in place.

PR #147, opened 2020-01-03
Replace ImmutableData with Blob

oetyng:
(For the full `safe-nd` refactor, see the demo implementation, at draft PR #145.)

- Adds new module `data`.
- Replaces `immutable_data.rs` with `blob.rs`, with renames but same functionality.
- Extracts tests out to `blob_tests.rs`.
- Renames all `IData` occurrences to `Chunk`.


src/data/blob_tests.rs
```diff
@@ -0,0 +1,89 @@
+// Copyright 2020 MaidSafe.net limited.
```
mrcnski, line 1: This file isn't really that big, and I think it could just be a sub-module in `data/blob.rs`: 

```rust
#[cfg(test)]
mod tests { 
... 
}
```

Not really an issue, but it would make the directory structure a bit simpler.

oetyng: It's for consistency, because `map.rs` and `sequence.rs` will have `map_tests.rs` and `sequence_tests.rs` in `data` module. Just not part of this PR.

Also I think there should be more tests there, looks a bit thin?

lionel-faber: Agreed that the types can have more tests.
Having a `tests` module in each map, sequence and blob seems to be a good idea.

Another option would be to have a `tests`module with separate files for each of the types. This is usually done when there are a _lot_ of tests. If that's the case then we can have `data/tests/blob.rs` and similar files for map and sequence. But if the tests are minimal then a sub-module should be fine.

mrcnski: Gotcha @oetyng and agreed with Lionel: I like `tests/blob.rs` more than `blob_tests.rs`. The former is what you'd more typically see in a Rust project.

lionel-faber:
@oetyng I'm unsure about the last commit that you've added renaming Blob to Chunk.
Is Chunk is being exposed as a publicly?

oetyng:
> @oetyng I'm unsure about the last commit that you've added renaming Blob to Chunk.
> Is Chunk is being exposed as a publicly?

It comes from this discussion:
https://github.com/maidsafe/safe-nd/pull/141#discussion_r362907565

What are the concerns wrt publicly exposed btw?

lionel-faber:
> What are the concerns 

The first thing that pops up in my mind that a private / public blob can be put directly to the network without self encryption. So it's technically not always a chunk.

oetyng:
> > What are the concerns
> The first thing that pops up in my mind that a private / public blob can be put directly to the network without self encryption. So it's technically not always a chunk.

You mean when it's smaller than the required size for self-encryption?
Well, it's more related to a chunk than a blob, if we define a blob as a set of chunks? A blob can be one or more chunks. A chunk can never be a blob since it cannot _be_ a set of chunks.

So, a chunk of data at size 1 of 1.

Not sure what alternative you propose @lionel1704 ?

@bochaco , what's your view on this?

bochaco:
Not sure I fully understand the concern, Blobs always use self-encryption, is this correct? if so it also can be a single chunk but still encapsulated in a Blob data type with all its features/APIs ?

If it's possible to put raw chunks also, then that's ok, we will base all our APIs and data structures on Blobs, Maps and Sequences though. I don't know if that's what you were after @lionel1704 ?

oetyng:
> Not sure I fully understand the concern, Blobs always use self-encryption, is this correct? if so it also can be a single chunk but still encapsulated in a Blob data type with all its features/APIs ?
> 
> If it's possible to put raw chunks also, then that's ok, we will base all our APIs and data structures on Blobs, Maps and Sequences though. I don't know if that's what you were after @lionel1704 ?

I'm not sure I do either. But there is a threshold where self-encryption is not applied. 
```
/// Holds the information that is required to recover the content of the encrypted file.  Depending
/// on the file size, this is held as a vector of `ChunkDetails`, or as raw data.
#[derive(Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Clone)]
pub enum DataMap {
    /// If the file is large enough (larger than 3072 bytes, 3 * MIN_CHUNK_SIZE), this algorithm
    /// holds the list of the file's chunks and corresponding hashes.
    Chunks(Vec<ChunkDetails>),
    /// Very small files (less than 3072 bytes, 3 * MIN_CHUNK_SIZE) are not split into chunks and
    /// are put in here in their entirety.
    Content(Vec<u8>),
    /// empty datamap
    None,
}
```
I just think that DataMap needs some massage to better capture the concept.

Its variants are Chunks and Content. It's easy to just call it for example

```
ChunkMap {
    Multiple(Vec<Details>),
    Single(Content),
}
```
or

```
DataMap {
    MultipleChunks(Vec<Details>),
    SingleChunk(Content),
}
```

A chunk means a part that you get when dividing a whole with a number, it does not mean "something self-encrypted" (which current `DataMap` enum above alludes to, without that actually being the inherent logic of the code - typical example of code and use case limitations and in the end accidental complexity, due to concept confusion).

So, when the divisor is 1, the chunk count is 1.. and in that special case there is naturally no self-encryption since it requires 3 parts. 

Very simple system.

lionel-faber:
> If it's possible to put raw chunks also

Yes @bochaco I was talking about raw chunks that can be put directly without going via self-encryption. So it's not a `DataMap::Chunks` or `DataMap::Content` it's just `Blob`.

oetyng:
>  I was talking about **raw chunks** that can be put directly without going via self-encryption. So it's not a `DataMap::Chunks` or `DataMap::Content` it's just `Blob`.

Its not just Blob though, and the above description is conflating different things.

- Blob consists of chunks.
- Chunks are either self encrypted or not.
- They are spread out in the network.

- If self-encrypted, the datamap, which has _details_ (i.e. metadata, **not** **chunks**), tells us both where locate and decrypt the data.
- If not self-encrypted, the datamap tells us where to locate the data.

-----

So, besides from not conflating these things, we simply modify the definition slightly like so:

- Blob consists of **one or more** chunks.

( optional clarification: 
- A Blob consisting of many chunks, is understood to have all its content split up on those chunks.
- A Blob with one chunk, is simply understood to contain all the content in the one chunk.
- A Blob split up over multiple chunks, can optionally undergo self-encryption.

--------------

Additionally, its important to differentiate between the network physical structure, and the code level abstractions.

In code, an enum Blob might have variants with metadata.. which is an _abstraction_.

Way too often, the network design suffers from influence of code level abstractions and decisions, and very rarely is the distinction clearly made - something that is necessary using the language and structure in the code.

bochaco:
Probably the confusion comes from what a blob means, but here we are talking about a `Blob` as the name of a data type we define, which is what @oetyng is describing above. The same for chunks, we are defining a Chunk to be a raw chunk/piece/blob of data which may/may-not be part of a Blob. So our implementations always work with `Blob`s, regardless if it contains one or more chunks, and I think we always will use self-encryption on it.

oetyng:
> and I think we always will use self-encryption on it.

@bochaco, this might not be quite possible the way things work (which I hinted to above, and you can read from the code comments there on `DataMap::Content`).
This is the answer I got from David the other day, when asking about minimum size:

> We have min size in self encryption, It is where the size would be less than the 3 hashes plus some buffer (3Kb) that will help. Less than that the pointers are bigger than the data. Idea is small files are part of a bigger group/directory etc. so the resulting data map (wiht unencrypted data) is in fact part of a bigger file and therefore is encrypted.

oetyng:
> The same for chunks, we are defining a Chunk to be a raw chunk/piece/blob of data which may/may-not be part of a Blob. So our implementations always work with `Blob`s, regardless if it contains one or more chunks

^This.

@lionel1704 , one way to think about it that can be helpful:

The distinction we make between `Blob` and `Chunk`, is conventional - since generally you could mix those words up. Our goal is naturally to define clear boundaries in our domain language, so that there is no confusion. 
We have a data type that we call `Blob`, and it consists of one or more `Chunks`. We reserve the word `Chunk`, for the concept it represents, and we reserve the word `Blob`, for the concept it represents. They are not interchangeable in this vocabulary. One thing is a data type, the other thing is a piece of data.

So, `Blob` is our system concept of a data type. The underlying chunks (now we're going into `Data Concept Proposal`, but that one is there for a reason; we need to have clear distinctions) do not pertain to any data type. They are just chunks spread out in the network.
We can access those chunks, via a data type abstraction. We might have more than one abstraction instance accessing _the same chunks_. That abstraction might be a `Blob`, might be a `Map`.. etc..

(If you look at the `Data Concept Proposal`, you will see how we try to separate chunks from the data types abstractions, and let chunks be the base for _all_ data in the system, and data types is just a framework through which we access them.)

So, the only reason a `Blob` might consist of a single chunk, is because there's a practical lower limit to self-encryption, as described above by David.
So, there is a base case when data size is too small for self encryption, and then the `Blob` is simply one chunk, instead of many.

lionel-faber:
> We have a data type that we call Blob, and it consists of one or more Chunks
This is right. But it's not always the case. The data types are used by both vaults and clients.
Vaults do not know about self encryption, as far as a vault is concerned, it receives a `Blob` which it stores / retrieves. The only Client <--> Network communication that happens is `GetBlob` and `PutBlob`.

Chunks / DataMaps etc. are purely a Client side thing, so we should introduce those names to the Vaults IMO.

oetyng:
> > We have a data type that we call Blob, and it consists of one or more Chunks
> This is right. But it's not always the case. The data types are used by both vaults and clients.
> Vaults do not know about self encryption, as far as a vault is concerned, it receives a `Blob` which it stores / retrieves. The only Client <--> Network communication that happens is `GetBlob` and `PutBlob`. 

There's something here to clear up. But I'll get back to this discussion, late tomorrow at earliest, otherwise on Friday.

> Chunks / DataMaps etc. are purely a Client side thing

I don't know about that.. a search for "chunk" in `safe_vault` gave 324 results in 21 files.

oetyng:
Due to the existing pervasive current and historical mixup in distinction of the file/object consisting of chunks, and the chunks themselves, with the words `ImmutableData`, `IData` and `Chunks`, this part of the proposal now waits a bit in line in favor of the other parts.

PR #149, opened 2020-01-09
chore/ci: add job-level if to condition on version change commits

mrcnski:
Note: This PR may change based on further replies to [this thread](https://github.community/t5/GitHub-Actions/Use-a-job-level-conditional-if-to-check-command-output/m-p/42911/highlight/false#M5170).

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -31,6 +31,7 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
+
       # Generate Cargo.lock, needed for the cache.
       - uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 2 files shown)


lionel-faber:
Introducing this change will make "Rebase and merge" compulsory since the latest commit (i.e. the commit that the action is being run against) might be a merge commit.
cc: @nbaksalyar

mrcnski:
Can you elaborate @lionel1704 ? This change should reduce LoC without making any functional changes.

lionel-faber:
If a PR is merged then the latest commit will be something like:
`Merge pull request #153 from joshuef/VersionBump0.7.2`
This would break the version change logic, yes?

mrcnski:
I think so, but we can check other commits as well. We can use an or-conditional to check if either the last or second-to-last commit is a version change. If you're happy with that I would be willing to make the change here, it should be fairly simple.

PR #154, opened 2020-01-17
Replace AppendOnlyData with Sequence

oetyng:
- Removes `append_only_data.rs`.
- Adds` sequence.rs` in `data` mod.
- Adds `access_control` in `data` mod.
- Adds rudimentary tests for `sequence`.
- Adds rudimentary tests for `access_control`.
- Adds `shared_types.rs`.
- Adds the `request` and `response` APIs for the `sequence` data type.


src/data/access_control.rs
```diff
@@ -0,0 +1,217 @@
...
+impl PublicUserAccess {
+    /// The ctor can be instantiated with
+    /// an access configuration.
+    pub fn new(config: BTreeMap<AccessType, bool>) -> Self {
+        // todo: filter out Queries
+        PublicUserAccess { config }
+    }
+
+    /// Returns `Some(true)` if `access` is allowed and `Some(false)` if it's not.
+    /// `None` means that `User::Anyone` permissions apply.
```
lionel-faber:
```suggestion
    /// `None` means that `User::Anyone` permissions may apply.
```

src/data/access_control.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub fn is_allowed(&self, access: AccessType) -> Option<bool> {
```
lionel-faber, line 123: Functions that start with `is...` usually return just a boolean

src/data/access_control.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// The list of users and their access configuration.
+    pub access_list: BTreeMap<User, PublicUserAccess>,
+    /// The expected index of the data at the time this grant config change is to become valid.
+    pub expected_data_version: u64,
+    /// The expected index of the owners at the time this grant config change is to become valid.
+    pub expected_owners_version: u64,
+}
+
+impl PublicAccessList {
+    fn is_allowed_(&self, user: &User, access: AccessType) -> Option<bool> {
```
lionel-faber, line 190: This function ends with a `_`?

src/data/access_control.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    /// Returns the list of users and their access configuration.
+    pub fn access_list(&self) -> &BTreeMap<User, PublicUserAccess> {
+        &self.access_list
+    }
+}
+
+impl AccessListTrait for PublicAccessList {
+    fn is_allowed(&self, user: &PublicKey, access: AccessType) -> bool {
+        self.is_allowed_(&User::Specific(*user), access)
```
lionel-faber, line 205: Ends with `_`?

src/data/access_control.rs
```diff
@@ -0,0 +1,217 @@
...
+impl From<PublicUserAccess> for UserAccess {
+    fn from(access: PublicUserAccess) -> Self {
+        UserAccess::Public(access)
+    }
+}
+
+/// The access configuration to Private data, for a User.
+#[derive(Clone, Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Hash, Debug)]
+pub struct PrivateUserAccess {
+    config: BTreeMap<AccessType, bool>,
```
lionel-faber: Why does this need need to mapped against a boolean? `User::Anyone` doesn't apply here so can it be a `BTreeMap<AccessType>` ?

src/data/sequence.rs
```diff
@@ -0,0 +1,727 @@
...
+pub struct DataEntry {
+    /// Data version
+    pub version: u64,
+    /// Data value
+    pub value: Vec<u8>,
+}
+
+impl DataEntry {
+    /// Returns a new instance of a data entry.
+    pub fn new(version: u64, value: Vec<u8>) -> Self {
```
lionel-faber: This constructor is not really needed since both the fields are `pub`

src/data/sequence.rs
```diff
@@ -0,0 +1,727 @@
...
+    C: AccessListTrait,
+    S: Copy,
+{
+    /// Returns true if the provided access type is allowed for the specific user (identified y their public key).
+    pub fn is_allowed(&self, user: PublicKey, access: AccessType) -> bool {
+        if let Some(owner) = self.owner_at(CURRENT_VERSION) {
+            if owner.public_key == user {
+                return true;
+            }
+        }
```
lionel-faber: self.is_owner() can be used here

src/data/sequence.rs
```diff
@@ -0,0 +1,727 @@
...
+    pub fn versions(&self) -> ExpectedVersions {
+        ExpectedVersions::new(
+            self.expected_data_version(),
+            self.expected_owners_version(),
+            self.expected_access_list_version(),
+        )
+    }
+
+    /// Returns the data shell - that is - everything except the Values themselves.
+    pub fn shell(&self, expected_data_version: impl Into<Version>) -> Result<Self> {
```
lionel-faber: I'm not sure what's the role of `expected_data_version` here?
Is this to fetch the data shell at a particular version of the data?

src/data/sequence.rs
```diff
@@ -0,0 +1,727 @@
...
+        let absolute_version = to_absolute_version(version, self.data.len())?;
+        self.data.get(absolute_version)
+    }
+
+    /// Return the current data entry (if it is present).
+    pub fn current_data_entry(&self) -> Option<DataEntry> {
+        match self.data.last() {
+            Some(value) => Some(DataEntry::new(self.data.len() as u64, value.to_vec())),
+            None => None,
+        }
```
lionel-faber: Perhaps it's better to use [map(..)](https://doc.rust-lang.org/std/option/enum.Option.html#method.map) here?

src/data/sequence.rs
```diff
@@ -0,0 +1,727 @@
...
+
+    /// Get owner at version.
+    pub fn owner_at(&self, version: impl Into<Version>) -> Option<&Owner> {
+        let version = to_absolute_version(version.into(), self.owners.len())?;
+        self.owners.get(version)
+    }
+
+    /// Returns history of all owners
+    pub fn owner_history(&self) -> Vec<Owner> {
+        self.owners.clone()
```
lionel-faber: Could this return a reference instead.
Do we need a copy of this vector anywhere?

src/data/sequence.rs
```diff
@@ -0,0 +1,727 @@
...
+    }
+
+    /// Get history of owners within the range of versions specified.
+    pub fn owner_history_range(&self, start: Version, end: Version) -> Option<Vec<Owner>> {
+        let range = to_absolute_range(start, end, self.owners.len())?;
+        Some(self.owners[range].iter().copied().collect())
```
lionel-faber: Some comment about the copy.
Also is there a specific reason why this returns an `Option`
If the range doesn't exist then it could return an empty vector instead?

lionel-faber: I just checked `to_absolute_range` and it returns Option there so it's fine to return an option here. :+1:

src/data/sequence.rs
```diff
@@ -0,0 +1,727 @@
...
+
+    /// Get access control at version.
+    pub fn access_list_at(&self, version: impl Into<Version>) -> Option<&C> {
+        let version = to_absolute_version(version.into(), self.access_list.len())?;
+        self.access_list.get(version)
+    }
+
+    /// Returns history of all access list states
+    pub fn access_list_history(&self) -> Vec<C> {
+        self.access_list.clone()
```
lionel-faber: Could we avoid this clone here?
If we need it then we can add it later but it's preferred that we pass references.

src/data/sequence.rs
```diff
@@ -0,0 +1,727 @@
...
+        }
+        if access_list.expected_owners_version() != self.expected_owners_version() {
+            return Err(Error::InvalidOwnersSuccessor(
+                self.expected_owners_version(),
+            ));
+        }
+        if self.expected_access_list_version() != expected_version {
+            return Err(Error::InvalidSuccessor(self.expected_access_list_version()));
+        }
+        self.access_list.push(access_list.clone()); // hmm... do we have to clone in situations like these?
```
lionel-faber: No the access_list argument shouldn't be a reference. That way the clone can be avoided.

lionel-faber (changes requested):
I've reviewed a part of the code. 
Left some comments so they could be addressed along with the other changes that are being discussed.

src/data/sequence.rs
oetyng: About copy: This was more or less the thing I was able to get through the compiler :)
It would seem that a reference to the slice would be most efficient, but I didn't manage to get that working. We could do a HO and look at it together to try figure it out?

src/data/access_control.rs
oetyng: Well maybe so, even though I wouldn't say that is quite relevant or necessarily puts a limit on this usage. This is a ternary operation now. The `is_allowed` question is still valid because the answer can be **Yes**, **No** and **Maybe**.

See, the answer to a question is not always binary :) And that's what's modeled here.
Feel free to suggest something else, but I think this is neatly captured in the code.

src/data/access_control.rs
oetyng: You can see the discussion here: https://github.com/maidsafe/safe-nd/pull/145#discussion_r362212825

And my conclusions at the end.
Relative to the faff it seemed like the lesser bad option, and I didn't quite get any push back on that then. Do you have something sleeker to suggest?

src/data/access_control.rs
oetyng: You can see the discussion here: https://github.com/maidsafe/safe-nd/pull/145#discussion_r362212825

And my conclusions at the end.
Relative to the faff it seemed like the lesser bad option, and I didn't quite get any push back on that then. Do you have something sleeker to suggest?

src/data/sequence.rs
oetyng: > Is this to fetch the data shell at a particular version of the data?

Yes. This logic has not changed from before. Only the name of the parameter.

src/data/sequence.rs
oetyng: Thing is, these places, when returning reference in original implementation, they clone at next level of usage (in `safe_vault`).
So I thought we might just as well do that from start.

src/data/sequence.rs
oetyng: Thing is, these places, when returning reference in original implementation, they clone at next level of usage (in `safe_vault`).
So I thought we might just as well do that from start.

src/data/access_control.rs
lionel-faber: Agreed that `Option<bool>` is absolutely needed here. But since that is the case, the function can be renamed something that makes sense, without the `is_` prefix. To keep it in line with Rust standards.

src/data/access_control.rs
lionel-faber: Well, IMO we should stick to the guidelines. It would certainly help in the long run when we might need to add more methods etc.
I don't have any sleeker ideas, will get back to you if I do.

src/data/sequence.rs
lionel-faber: I see. But "expected" doesn't seem to fit this context.
I think it's better to change it to just "version" so I can get the data shell at version 'n'. It would need changes in the logic  which should be simple.

src/data/sequence.rs
lionel-faber: Sure! I'll pull this code in locally and give it a try as well. Happy for a HO too.

src/data/sequence.rs
lionel-faber: The thing is that by keeping `clone()` in the upper layers we have absolute control of its necessity. Otherwise, clippy will complain. By cloning it here, we risk losing track of the number of clones happening.

src/data/access_control.rs
oetyng: I would say it makes perfect sense though.
Are there rust standards saying that a question (that what "is_" denotes) cannot have answers Yes, No and Maybe?
It seems like a very strange an strict standard in that case. There must be some level of problem solving where the tools are used in ways where they solve problems, not only in the narrow scopes that someone before thought of.

This level seems more like a question of design and modelling, than a language standards area. But happy to hear about that standard, I haven't actually seen the list of those.

src/data/sequence.rs
oetyng: If you are talking of the shell method parameter, then that's a simple rename.

The other methods in the highlighted code, deal with that version that we by convention call expected version.

src/data/sequence.rs
oetyng: OK, I had no complaints from clippy in safe_vault about it.
It got more complicated at that level though (it seemed). When I did this I mostly went for the first seemingly OK-ish thing that compiled. Didn't want to spend too much time on details that are easily sorted out by a rust coder.

In case you check out the safe_vault PR you can see the usage there, and if you know how it should be dealt with there.
Otherwise I'll have another look at it in a bit.

src/data/access_control.rs
lionel-faber: Yea there doesn't seem to be a fixed set of standards as such. [I found this](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#predicates) which seems relevant and makes sense, i.e. variables / functions that represent / return `bool` are prefixed by a short question word.
To think of it a bit differently,
We can do things like `is_none()` on `Option`.
`app.get_permission().is_none()` seems to make more sense than `app.is_allowed().is_none()` IMO.
Happy to see what others think as well.

src/data/sequence.rs
lionel-faber: Yea we do. It's just that "expected version" doesn't seem to fit in a get operation.
We'll need to define some stuff clearly. 
Eg. Do versions start from 0 or 1?
When we insert a new value for the first time, is it version 0 or version 1?
I think it's worth getting this clarified and do it right this time.

src/data/sequence.rs
lionel-faber: It's okay to `.clone()` in safe_vault and other layers. As I mentioned before, by restricting it at the lower layer, we avoid the always-clone that would happen in this case. i.e. we wouldn't be aware of it. I've seen places where two getters are available for both the clone and reference. That way the user of the API is aware of the extra memory usage. I'll take a look at the usage in safe_vault, thanks!

src/data/sequence.rs
oetyng: > "expected version" doesn't seem to fit in a get operation

It depends. As a parameter, where you are asking for some actual data, then yes, it should just be Version. Definitely.
If you are asking for the expected version (i.e. what version is expected?) Then.. you do a GET and the returned result are the "expected versions".

> We'll need to define some stuff clearly.
> Eg. Do versions start from 0 or 1?

What you request is part of the previous PR I did ([as seen here](https://github.com/maidsafe/safe-nd/pull/126)), when we introduced current and expected version convention. All that is included here. Versions always start at 0.

src/data/access_control.rs
oetyng: Yeah, it says this
> Simple boolean predicates should be prefixed with is_ or another short question word, e.g., is_empty.

That's standard pretty much anywhere I would say.
But that practice _does not_ say that a more rich question, which can return the equivalent of a "maybe" answer, is not allowed or would be a bad practice. It simply does not regard that possibility, which IMO is just a lack of foreseeing. It's not a dictate or an opinion on it.
Coding stadards are full of this, and that's where the new standards are born, by those who fill in the gaps.

To the suggestion of yours, I would not consider it a good option because it is too technical. I'm very much convinced that good and clear code uses the naming to explain intention and domain meaning, not to describe the technical ways we solve it.
So to ask if it is "None", is simply not helpful.
The `.is_none()` method is not used on this return value, it is simply used in a match clause, where it is terse and simple.

We want to know _if it is allowed_, and the answer can be yes, no and maybe, in which case we dig a bit deeper to find the answer.

### To conclude
While there are probably other solutions as well, of which some could be better. I think this is _far too small of an issue_ to spend a lot of time on now. It's one of those things that could be added to a "code style doc" for whenever we have a joint session in the team to prioritize important code style questions and come to decisions.

lionel-faber: Agreed :+1:

src/data/sequence.rs
lionel-faber: > the returned result are the "expected versions".
Yes it makes sense for a result. But as a paramter is what I'm not sure of.

This function is to return the shell of the data. So `data.shell(0)` is get shell at version 0 `data.shell(1)` is get shell at version 1. Which makes sense. 

But right now it's implemented as `data.shell(1)` which interprets as get the shell of the data when the next expected version is 1. Which might be confusing right?

src/data/sequence.rs
```diff
@@ -0,0 +1,603 @@
...
+    }
+
+    /// Returns history of all access list states
+    pub fn public_access_list_history(&self) -> Result<Vec<PublicAccessList>> {
+        use Sequence::*;
+        let result = match self {
+            Public(data) => Some(data.access_list_history()),
+            _ => return Err(Error::InvalidOperation),
+        };
+        result.ok_or(Error::NoSuchEntry)
```
lionel-faber: I think `if let` would be a better fit here.
```
if let Sequence::Public(data) = self {
    Ok(data.access_list_history())
} else {
    Err(Error::InvalidOperation)
}
```
Also I'm not sure what kind of error we want to return here. If the type is private it will return `InvalidOperation` but in the above code result is always `Some` so `NoSuchEntry` will never be returned here. Not sure what was intended here.

lionel-faber: I see a similar pattern in some other functions too.

src/data/access_control.rs
lionel-faber: This can be done in a follow-up PR.

src/data/access_control.rs
lionel-faber: Can be done in a follow up PR.

lionel-faber:
I pulled this code locally and addressed some of the concerns with cloning in certain places.
[See the diff](https://github.com/oetyng/safe-nd/compare/replace-ad-with-sequence...lionel1704:replace-ad-with-sequence?expand=1) for reference. Apart from this there is only [one older comment](https://github.com/maidsafe/safe-nd/pull/154#discussion_r371728411) that needs some addressing / discussion and a [new observation](https://github.com/maidsafe/safe-nd/pull/154#discussion_r372797717) that I just made which needs addressing. With those two out of the way we can merge this PR.

EDIT: There's also a change proposed by Adam in #82 that would be nice to have in the new types from the start but it can be in a follow-up PR as well.

src/data/sequence.rs
oetyng: Yeah absolutely, that must be some remnant, line 538 has it correct, the parameter is named `version` there.

It looked like it's just the name of the parameter that is wrong, and that the logic is as it should (you'd pass in 0 and get at version 0).

But I see it is using old logic there, and with the rename, that logic was basically expecting the "expected version". So, will update name and logic.

src/data/sequence.rs
oetyng: Great observation. I basically tried with first pattern that came to mind on those places. This is naturally much better.
Are there more places to fix now after the latest commits?

src/data/sequence.rs
```diff
@@ -0,0 +1,601 @@
...
+    }
+
+    /// Returns the access list of a public instance at a specific version.
+    pub fn public_access_list_at(&self, version: impl Into<Version>) -> Result<&PublicAccessList> {
+        use Sequence::*;
+        let access_list = match self {
+            Public(data) => data.access_list_at(version),
+            _ => return Err(Error::InvalidOperation),
+        };
+        access_list.ok_or(Error::NoSuchEntry)
```
lionel-faber: Can use `if let ...` here.

src/data/sequence.rs
```diff
@@ -0,0 +1,601 @@
...
+    pub fn private_access_list_at(
+        &self,
+        version: impl Into<Version>,
+    ) -> Result<&PrivateAccessList> {
+        use Sequence::*;
+        let access_list = match self {
+            Private(data) => data.access_list_at(version),
+            _ => return Err(Error::InvalidOperation),
+        };
+        access_list.ok_or(Error::NoSuchEntry)
```
lionel-faber: Can use `if let ...` here.

src/data/sequence.rs
```diff
@@ -0,0 +1,601 @@
...
+        &self,
+        start: Version,
+        end: Version,
+    ) -> Result<Vec<PublicAccessList>> {
+        use Sequence::*;
+        let result = match self {
+            Public(data) => data.access_list_history_range(start, end),
+            _ => return Err(Error::InvalidOperation),
+        };
+        result.ok_or(Error::NoSuchEntry)
```
lionel-faber: Can use `if let ....` here too.

src/data/sequence.rs
```diff
@@ -0,0 +1,601 @@
...
+        &self,
+        start: Version,
+        end: Version,
+    ) -> Result<Vec<PrivateAccessList>> {
+        use Sequence::*;
+        let result = match self {
+            Private(data) => data.access_list_history_range(start, end),
+            _ => return Err(Error::InvalidOperation),
+        };
+        result.ok_or(Error::NoSuchEntry)
```
lionel-faber: Here as well.

src/data/sequence.rs
```diff
@@ -0,0 +1,601 @@
...
+        state_dispatch!(self, ref mut state => state.set_owner(owner, expected_version))
+    }
+
+    /// Sets a new access list of a private instance.
+    pub fn set_private_access_list(
+        &mut self,
+        access_list: PrivateAccessList,
+        expected_version: u64,
+    ) -> Result<()> {
+        use Sequence::*;
```
lionel-faber: When needing only one variant, we can avoid wildcard imports (`::*`)

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,341 @@
...
+        expected_data_version: 0,
+        expected_owners_version: 0,
+    };
+    // Set the first access_list with correct ExpectedVersions - should pass.
+    let res = data.set_access_list(access_list, 0);
+
+    match res {
+        Ok(()) => (),
+        Err(x) => panic!("Unexpected error: {:?}", x),
+    }
```
lionel-faber: Can use `unwrap!` macro in tests.

`unwrap!(data.set_access_list(access_list, 0))`. Basically does the same thing.

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    // Verify that the access_list are part of the history.
+    assert_eq!(
+        unwrap!(data.access_list_history_range(Version::FromStart(0), Version::FromEnd(0),)).len(),
```
lionel-faber: Trailing `,`

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,341 @@
...
+        expected_data_version: 64,
+        expected_owners_version: 0,
+    };
+    // Set access_list with incorrect ExpectedVersions - should fail.
+    let res = data.set_access_list(access_list, 1);
+
+    match res {
+        Err(_) => (),
+        Ok(()) => panic!("Unexpected Ok(()) result"),
+    }
```
lionel-faber: Can use `unwrap!`

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,341 @@
...
+            expected_data_version: 0,
+            expected_access_list_version: 0,
+        },
+        0,
+    );
+
+    match res {
+        Ok(()) => (),
+        Err(x) => panic!("Unexpected error: {:?}", x),
+    }
```
lionel-faber: Can use `unwrap!()`

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,341 @@
...
+            expected_data_version: 64,
+            expected_access_list_version: 0,
+        },
+        1,
+    );
+
+    match res {
+        Err(_) => (),
+        Ok(()) => panic!("Unexpected Ok(()) result"),
+    }
```
lionel-faber: Can use `unwrap!`

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,341 @@
...
+    );
+
+    let _ = data.set_owner(
+        Owner {
+            public_key: owner_pk1,
+            expected_data_version: 0,
+            expected_access_list_version: 0,
+        },
+        1,
+    );
```
lionel-faber: Result shouldn't be ignored here. Use `unwrap!`

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,341 @@
...
+    assert_sequence_read_permitted(&data, &public_key_0, true);
+    assert_sequence_read_permitted(&data, &public_key_1, true);
+
+    // with access_list
+    let mut access_list = PublicAccessList {
+        access_list: BTreeMap::new(),
+        expected_data_version: 0,
+        expected_owners_version: 1,
+    };
+    let mut set = BTreeMap::new();
```
lionel-faber: Maybe a more meaningful name here?

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,341 @@
...
+    assert_sequence_read_permitted(&data, &public_key_0, true);
+    assert_sequence_read_permitted(&data, &public_key_1, false);
+
+    // with access
+    let mut access_list = PrivateAccessList {
+        access_list: BTreeMap::new(),
+        expected_data_version: 0,
+        expected_owners_version: 1,
+    };
+    let mut set = BTreeSet::new();
```
lionel-faber: Here too. Maybe something more meaningful.

src/errors.rs
```diff
@@ -53,8 +52,8 @@ pub enum Error {
     InvalidEntryActions(BTreeMap<Vec<u8>, EntryError>),
     /// Key does not exist
     NoSuchKey,
-    /// The key(s) of the entry or entries contained in this error already exist
-    KeysExist(ADataEntries),
+    // /// The key(s) of the entry or entries contained in this error already exist
+    // KeysExist(ADataEntries),
```
lionel-faber: Can be removed fully?

src/shared_types.rs
```diff
@@ -0,0 +1,252 @@
...
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, Debug)]
+pub enum User {
+    /// The category representing any user.
+    Anyone,
+    /// The category representing a specific user.
+    Specific(PublicKey),
+}
+
+/// The current version is defined as the last entry in the vector.
+pub const CURRENT_VERSION: Version = Version::FromEnd(1); // Shouldn't last entry be accessed with 0 ?
```
lionel-faber: > // Shouldn't last entry be accessed with 0 ?

Yes ?

src/shared_types.rs
```diff
@@ -0,0 +1,252 @@
...
+
+///
```
lionel-faber: Missing doc comment?

src/shared_types.rs
```diff
@@ -0,0 +1,252 @@
...
+/// Set of data, owners, permissions versions.
+#[derive(Copy, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+pub struct ExpectedVersions {
+    expected_data_version: u64,
+    expected_owners_version: u64,
+    expected_access_list_version: u64,
+}
+
+impl ExpectedVersions {
+    /// ctor
```
lionel-faber: Perhaps something more meaningful here.

src/shared_types.rs
```diff
@@ -0,0 +1,252 @@
...
+        ///
+        tag: u64,
+    },
+    ///
+    Private {
+        ///
+        name: XorName,
+        ///
+        tag: u64,
+    },
```
lionel-faber: Leaving a comment just to make a note of this.

lionel-faber (commented):
Sorry @oetyng. I didn't look at the tests when reviewing this previously.
I'll be going through these changes to address some of the comments raised previously.
I can address these comments in the same follow up PR.
I'm okay with getting this merged as soon after rebase and it passing CI.

src/data/sequence.rs
```diff
@@ -0,0 +1,604 @@
...
+    // could represent an individual user, or a group of users, depending on the `PublicKey` type.
+    owners: Vec<Owner>,
+}
+
+/// Common methods both `Sequence` scopes (Public | Private).
+impl<C> SequenceBase<C>
+where
+    C: AccessListTrait,
+{
+    /// Returns true if the provided access type is allowed for the specific user (identified y their public key).
```
lionel-faber:
```suggestion
    /// Returns true if the provided access type is allowed for the specific user (identified by their public key).
```

src/data/tests/access_control.rs
```diff
@@ -0,0 +1,326 @@
...
+    unwrap!(data.set_owner(
+        Owner {
+            public_key: owner_pk1,
+            expected_data_version: 0,
+            expected_access_list_version: 0,
+        },
+        1,
+    ));
+
+    dbg!(unwrap!(data.shell(None)).owner_history());
```
lionel-faber:
```suggestion
```

src/data/tests/mod.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2020 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+#[cfg(test)]
```
lionel-faber: `#[cfg(test)]` should be in `data/mod.rs` above `mod tests`

src/shared_types.rs
```diff
@@ -0,0 +1,224 @@
...
+        Version::FromEnd(version) => count.checked_sub(version as usize),
+    }
+}
+
+pub fn to_absolute_range(start: Version, end: Version, count: usize) -> Option<Range<usize>> {
+    let start = to_absolute_version(start, count)?;
+    let end = to_absolute_version(end, count)?;
+
+    if start <= end {
+        Some(start..end) // `end` is exclusive in Range
```
lionel-faber:
```suggestion
        Some(start..end)
```
Since `+1` is removed, the comment is not required.

src/data/tests/mod.rs
oetyng: fixed

src/data/tests/access_control.rs
oetyng: fixed

src/data/sequence.rs
oetyng: fixed

src/shared_types.rs
oetyng: fixed

oetyng:
> I'm okay with getting this merged as soon after rebase and it passing CI.

The problem with CI has been there for a while. It never finishes.

lionel-faber:
Looks like CI isn't completing since the test fails on Ubuntu. The windows / mac builds are automatically cancelled after that.
Could you check that out?

oetyng:
This is (roughly) superseded by PR #163.
A new take on the peripheral changes included in this PR needs to be done after CRDT data types are in place.

PR #173, opened 2020-06-26
At2 get replica keys

joshuef:


src/request/login_packet.rs
```diff
@@ -23,7 +23,7 @@ pub enum LoginPacketRequest {
     Create {
         /// The LoginPacket itself.
         login_packet: LoginPacket,
-        /// Debit-proof for storing the LoginPacket
+        /// Debit-proof for storing the LoginPacket - Optional fir IdataOps
```
Yoga07, line 26: We can remove `- Optional for IdataOps` from there and add it [here](https://github.com/maidsafe/safe-nd/pull/173/files#diff-e0530a20ff10cf0f1134fca1113e93b3R22).

Yoga07 (commented):
Just a minor change

Yoga07 (approved):

PR #177, opened 2020-07-01
refactor/request: introduce read/write

oetyng:
- Replaces Get/Mutation with Read/Write,
 as a first step of Request structure refactor.
 - MutableData not included as it is still due for CRDT refactor.


Yoga07:
Superseded by [maidsafe/farming](https://github.com/maidsafe/safe-nd/tree/farming) branch

PR #178, opened 2020-07-01
Separate read write

oetyng:
Follow up PR to #177 

Completes the refactor of `Request` with separation of reads and writes.


oetyng:
outdated

PR #181, opened 2020-07-13
chore(ci): Add workflows to enable continuous delivery

joshuef:


bochaco (approved):

.github/workflows/bump_version.yml
```diff
@@ -0,0 +1,33 @@
...
+env:
+  NODE_ENV: 'development'
+  GITHUB_TOKEN: ${{ secrets.PERSONAL_GITHUB_TOKEN }}
+
+jobs:
+  bump:
+    runs-on: ubuntu-latest
+    if: "!startsWith(github.event.head_commit.message, 'chore(release):')"
+    steps:
+      - uses: actions/checkout@v1
```
S-Coyle, line 18: There's a v2 of the checkout action available now, it offers some decent [improvements](https://github.com/actions/checkout#whats-new) over v1, may as well switch to that.

Also, I noticed recently that tags weren't being produced in some repos when the Version change PRs were going through CI and it seemed to be because we needed to checkout with `fetch-depth: '0'` to be sure to retrieve all commits to look for the semver commit message. I added with something like:
```
     - uses: actions/checkout@v2
        with:
          fetch-depth: '0'
```
Might be worth adding here just in case the same thing ends up happening again

S-Coyle: Ah just moved onto reviewing commitlint.yml & I see you have both v2 & the fetch-depth there - should also add into bump_version.yml then?

S-Coyle:
Not a big deal - a few `no new line at end of file` warnings in there

joshuef:
working on a GHA just fyis

joshuef:
Superseded by #185

PR #183, opened 2020-07-14
feat(Sequence): support Policies change concurrently with data ops

bochaco:


bochaco:
Closing this for now, we are going for the approach implemented by PR #186

PR #186, opened 2020-07-20
feat(Sequence): support Policies change concurrently with data ops

bochaco:


src/request/sdata.rs
```diff
@@ -46,25 +46,21 @@ pub enum SDataRequest {
...
-    /// Get current owner.
-    GetOwner(SDataAddress),
-    /// Set new permissions for public Sequence.
-    MutatePubPermissions(SDataMutationOperation<SDataPubPermissions>),
-    /// Set new permissions for private Sequence.
-    MutatePrivPermissions(SDataMutationOperation<SDataPrivPermissions>),
-    /// Add a new `owners` entry. Only the current owner(s) can perform this action.
-    MutateOwner(SDataMutationOperation<SDataOwner>),
+    /// Set new policy for public Sequence.
+    MutatePubPolicy(SDataDataMutationOp<SDataPubPolicy>),
```
Scorch-Dev: Could you explain why the `MutatePubPolicy` and `MutatePrivPolicy` variants are backed by `SDataDataMutationOp<T>`? Should  they be `SDataPolicyMutationOp<T>` instead, or am I missing something?

bochaco: It could be yes, it's just that since it's WIP and still didn't finish the part of defining the policy ops it's not properly reflected there.

Scorch-Dev: That makes sense, thanks for the clarification 👍

bochaco:
Superseded by PR #216

PR #201, opened 2020-08-04
refactor(rewards): claim counter -> get account id

oetyng:
- Removes granular counter, and simplifies flow to only send
account id for reward payout at node relocation.
- Also simplifies storecost to only be num bytes.


Yoga07 (approved):
Nice one :+1:

PR #211, opened 2020-08-25
Prevent cloning of messages unless absolutely required

lionel-faber:


src/messaging/mod.rs
```diff
@@ -88,10 +88,8 @@ impl MsgEnvelope {
     /// The proxy would first sign the MsgEnvelope,
     /// and then call this method to add itself
     /// (public key + the signature) to the envelope.
-    pub fn with_proxy(&self, proxy: MsgSender) -> MsgEnvelope {
-        let mut clone = self.clone();
-        clone.proxies.push(proxy);
-        clone
+    pub fn add_proxy(&mut self, proxy: MsgSender) {
+        self.proxies.push(proxy);
```
lionel-faber, line 92: An alternative to this would be to make the `with_proxy` consume `self` and return a new message so that we don't keep two copies of the MsgEnvelope

Yoga07 (approved):

PR #249, opened 2020-10-09
fix/sequence: fix in_range getter for sequence type

Yoga07:


bochaco (dismissed):

src/sequence/seq_crdt.rs
```diff
@@ -327,7 +327,7 @@ where
 
         self.current_lseq().map(|lseq| {
             lseq.iter()
-                .take(end_index - 1)
+                .take(end_index)
```
bochaco: Perhaps it's better to rename the var from `end_index` to `num_items` or something like it

bochaco (approved):

PR #253, opened 2020-10-20
Use abstracted keypair, and create SecretKey abstraction over ed/bls key types

joshuef:


src/keys/mod.rs
```diff
@@ -399,6 +437,23 @@ pub struct BlsKeypairShare {
     pub public_key_set: threshold_crypto::PublicKeySet,
 }
 
+/// Wrapper for different secret key types.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum SecretKey {
+    /// Ed25519 keypair.
```
Yoga07: Better to call them `SecretKey` rather than `keypair`?

Yoga07 (dismissed):

Yoga07 (approved):

PR #255, opened 2020-10-22
Feat sk from threshold sk + keys refactor

joshuef:


src/keys/keypair.rs
```diff
@@ -0,0 +1,241 @@
...
+use ed25519_dalek::Signer;
+use rand::{CryptoRng, Rng};
+use serde::{Deserialize, Serialize};
+use std::fmt::{self, Debug, Formatter};
+use threshold_crypto::{self, serde_impl::SerdeSecret};
+use unwrap::unwrap;
+
+/// Wrapper for different keypair types.
+#[derive(Serialize, Deserialize)]
+pub enum Keypair {
```
dirvine, line 28: Should this one be `Keypairs` plural?

src/keys/keypair.rs
```diff
@@ -0,0 +1,241 @@
...
+
+use crate::{Error, Result};
+use crate::{PublicKey, SecretKey, Signature, SignatureShare};
+
+use ed25519_dalek::Signer;
+use rand::{CryptoRng, Rng};
+use serde::{Deserialize, Serialize};
+use std::fmt::{self, Debug, Formatter};
+use threshold_crypto::{self, serde_impl::SerdeSecret};
+use unwrap::unwrap;
```
dirvine, line 24: we should not use this, it's defunct now.

src/keys/keypair.rs
```diff
@@ -0,0 +1,241 @@
...
+    Bls(BlsKeypair),
+    /// BLS keypair share.
+    BlsShare(BlsKeypairShare),
+}
+
+// Need to manually implement this due to a missing impl in `Ed25519::Keypair`.
+impl Clone for Keypair {
+    fn clone(&self) -> Self {
+        match self {
+            Self::Ed25519(keypair) => Self::Ed25519(unwrap!(ed25519_dalek::Keypair::from_bytes(
```
dirvine: This is dangerous, both Clone and unwrap!

src/keys/keypair.rs
joshuef: Not yet. It's just a wrapper around the various keypair types.

This isn't renamed, so I'd ideally not want to touch this here if we're going to refactor this all shortly anywho

src/keys/keypair.rs
joshuef: So I've taken the olde impl we already had here.

I know we don't want `unwrap` ideally... but in a `clone` or `from` that can't have a `Result` type what's the best way?

Is it actually not to `clone`, but some custom func? (I guess just not use the trait?)

src/keys/keypair.rs
joshuef: see next comment :+1:

src/keys/keypair.rs
```diff
@@ -0,0 +1,241 @@
...
+
+impl From<SerdeSecret<threshold_crypto::SecretKey>> for Keypair {
+    fn from(sk: SerdeSecret<threshold_crypto::SecretKey>) -> Self {
+        let public = sk.public_key();
+        let keypair = BlsKeypair { secret: sk, public };
+        Self::Bls(keypair)
+    }
+}
+
+impl From<ed25519_dalek::SecretKey> for Keypair {
```
dirvine, line 170: This one is unusual perhaps as Dalek holds a Keypair type that already has secret and public in it?

joshuef: It does, but it's not our abstracted keypair. Which can be either bls/share or ed

joshuef:
closing as this is more than likely superseded by PublicId removal

PR #256, opened 2020-10-26
Feat remove public ids

joshuef:


src/keys/keypair.rs
```diff
@@ -0,0 +1,241 @@
...
+
+use crate::{Error, Result};
+use crate::{PublicKey, SecretKey, Signature, SignatureShare};
+
+use ed25519_dalek::Signer;
+use rand::{CryptoRng, Rng};
+use serde::{Deserialize, Serialize};
+use std::fmt::{self, Debug, Formatter};
+use threshold_crypto::{self, serde_impl::SerdeSecret};
+use unwrap::unwrap;
```
dirvine: We still need to remove this.

src/keys/keypair.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// Ed25519 keypair.
+    Ed25519(ed25519_dalek::Keypair),
+    /// BLS keypair.
+    Bls(BlsKeypair),
+    /// BLS keypair share.
+    BlsShare(BlsKeypairShare),
+}
+
+// Need to manually implement this due to a missing impl in `Ed25519::Keypair`.
+impl Clone for Keypair {
```
dirvine: We likely don't need Clone and the unwrap if we use Arc (for now anyway)

src/keys/keypair.rs
```diff
@@ -0,0 +1,241 @@
...
+        ]
+    }
+
+    fn gen_keys() -> Vec<PublicKey> {
+        gen_keypairs().iter().map(PublicKey::from).collect()
+    }
+
+    // Test serialising and deserialising key pairs.
+    #[test]
+    fn serialisation_key_pair() {
```
dirvine: if we make this a `Result<>` we can use a `?` and not unwrap (we need to kill that crate anyway. If we need unwrap we can use the std `.unwrap("comment")`

src/keys/keypair.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    /// Signs with the underlying keypair.
+    pub fn sign(&self, data: &[u8]) -> Signature {
+        match self {
+            Self::Ed25519(keypair) => Signature::Ed25519(keypair.sign(&data)),
+            Self::Bls(keypair) => Signature::Bls(keypair.secret.sign(data)),
+            Self::BlsShare(keypair) => {
+                let index = keypair.index;
+                let share = keypair.secret.sign(data);
+                Signature::BlsShare(SignatureShare { index, share })
```
Yoga07:
```suggestion
                Signature::BlsShare(SignatureShare { index: keypair.index, share: keypair.secret.sign(data) })
```
This way we can remove the unnecessary init of two vars above. We don't seem to use them elsewhere.

src/keys/keypair.rs
```diff
@@ -0,0 +1,246 @@
...
+            public,
+        };
+        Self::Bls(keypair)
+    }
+}
+
+impl From<SerdeSecret<threshold_crypto::SecretKey>> for Keypair {
+    fn from(sk: SerdeSecret<threshold_crypto::SecretKey>) -> Self {
+        let public = sk.public_key();
+        let keypair = BlsKeypair { secret: sk, public };
```
Yoga07: Same as above. Don't seem to use `public` and `keypair` elsewhere.

src/keys/keypair.rs
```diff
@@ -0,0 +1,246 @@
...
+        Self::Bls(keypair)
+    }
+}
+
+impl From<&SerdeSecret<threshold_crypto::SecretKey>> for Keypair {
+    fn from(sk: &SerdeSecret<threshold_crypto::SecretKey>) -> Self {
+        let public = sk.public_key();
```
Yoga07: Same as above.

Yoga07 (dismissed):
Suggested a few code style changes. LGTM otherwise :+1:

Yoga07 (approved):

PR #259, opened 2020-10-28
tests: add proptests for Sequence data

joshuef:


dirvine (dismissed):

src/sequence/mod.rs
```diff
@@ -719,22 +723,162 @@ mod tests {
...
+
+    proptest! {
+        #[test]
+        fn proptest_seq_doesnt_crash_with_random_data(s in "\\PC*".prop_map(|s| s.into_bytes())) {
+            let actor1 = gen_public_key();
+            let sequence_name = XorName::random();
+
+            let sdata_tag = 43_001u64;
+
+            // Instantiate the same Sequence on two replicas with two diff actors
```
oetyng: Is the "two diff actors" an old comment, or am I missing something? Seems to be one and the same?
(Same thing in the other tests)

src/sequence/mod.rs
```diff
@@ -719,22 +723,162 @@ mod tests {
...
+            // insert our data at replicas
+            for data in dataset {
+                let data = data.into_bytes();
+
+                // first we generate standard data ops
+                for replica in &mut replicas {
+                    let append_op = replica.append(data.clone())?;
+                    let mut ops = vec![];
+                    ops.push(append_op);
+                    ops.shuffle(&mut OsRng);
```
oetyng: I think the intention is to build up ops in the ops vec, right? In that case the instantiation just needs to be moved out of the `for replica` loop.

src/sequence/mod.rs
```diff
@@ -719,22 +723,162 @@ mod tests {
...
+            replica2.apply_public_policy_op(owner_op)?;
+
+            // Append an item on replicas
+            let append_op1 = replica1.append(s.clone())?;
+            let append_op2 = replica2.append(s)?;
+
+            replica1.apply_data_op(append_op1)?;
+            replica2.apply_data_op(append_op2)?;
+
+            verify_data_convergence(vec![replica1, replica2], 1);
```
bochaco: I think this test should fail? we have appended the same bytes to both replicas, but they are different items, perhaps we can use different bytes for appending to each replica so we can easily differentiate them?.
Also, we haven't applied the operation from replica1 to replica2, and vice versa, which should be the normal flow and then the convergence check should expect two items on each replica.

src/sequence/mod.rs
```diff
@@ -719,22 +723,162 @@ mod tests {
...
-                assert_eq!(r0_entry, r.get(index));
+            let r0_entry = &replicas[0].get(index);
+            for r in &replicas {
+                assert_eq!(r0_entry, &r.get(index));
             }
         }
     }
+
+    /// Generate a vec of Sequence replicas of some length
+    fn generate_replicas(quantity: i32, xorname: XorName, tag: u64) -> Result<Vec<Sequence>> {
```
bochaco: Not sure it makes sense to generate replicas all with same xorname and tag, I think we can either generate those values here randomly, or have this function to receive `Vec<(XorName, u64)>` as arg. instead, and it generates one replica per vec item.

src/sequence/mod.rs
```diff
@@ -719,22 +723,162 @@ mod tests {
...
+                let data = data.into_bytes();
+                // Append an item on replica1
+                let append_op1 = replica1.append(data.clone())?;
+                let append_op2 = replica2.append(data)?;
+
+                replica1.apply_data_op(append_op1)?;
+                replica2.apply_data_op(append_op2)?;
+            }
+
+            verify_data_convergence(vec![replica1, replica2], dataset_length);
```
bochaco: Same comment as in previous test about this check, we are having a false positive testing convergence here if I'm correct.

src/sequence/mod.rs
```diff
@@ -719,22 +723,162 @@ mod tests {
...
+
+                for replica in &mut replicas {
+                    let append_op = replica.append(data.clone())?;
+
+                    replica.apply_data_op(append_op)?;
+
+                }
+            }
+
+            verify_data_convergence(replicas, dataset_length);
```
bochaco: Same comment as previous tests.

src/sequence/mod.rs
```diff
@@ -719,22 +723,162 @@ mod tests {
...
+                    for op in ops {
+
+                        replica.apply_data_op(op)?;
+                    }
+
+                }
+
+            }
+
+            verify_data_convergence(replicas, dataset_length);
```
bochaco: Same comment as previous tests.

bochaco (changes requested):
As an enhancement, I think for things like `proptest_converge_with_shuffled_data_set_across_arbitrary_number_of_replicas` we should start creating our custom strategies, e.g. in this case an strategy which gives you a Vec of operations (eventually this can contain not only append but remove ops too).

src/sequence/mod.rs
joshuef: ach, you're correct good spot :+1:

src/sequence/mod.rs
joshuef: At the moment, I was just intending to test shuffling ops for _that replica_ only. 

If we move it out, we'd get an `ops` of all ops across all replicas, which would fail when trying to apply ops to another replica.

`Test failed: Data operation depends on a different replica's state than the current;`

joshuef: This may be me misunderstanding how we should be applying this stuff though (looking at gabs other comments), so perhaps what you're suggesting _should_ work...

src/sequence/mod.rs
joshuef: But isn't it a replica of a given piece of data in the network? Like, each replica will be managing data of xorname `X` and tag: `64`, that's how we know it's the same data... Otherwise what's to converge upon, it could be a completely different piece of data, no?

src/sequence/mod.rs
joshuef: Grand, have this set up now, w/ one op set generated for one replica, and then that is shuffled + applied for each replica thereafter :+1:

src/sequence/mod.rs
joshuef: Aha, I was thinking our `verify_convergence` was doing more than it is.

src/sequence/mod.rs
bochaco: yeah, you are right, i had to refresh my memory with this :)

src/sequence/mod.rs
```diff
@@ -719,22 +724,368 @@ mod tests {
...
+            //
+            // Q: how to expand this across many replicas and simulate dropping messages? Is that wortwhile here at the data itself...?
+            verify_data_convergence(replicas, dataset_length);
+        }
+
+        #[test]
+        fn proptest_converge_with_shuffled_ops_including_bad_ops_which_error_and_are_not_applied(
+            dataset in generate_dataset(1000),
+            // should be same number as dataset
+            bogus_dataset in generate_dataset(1000),
```
bochaco, line 1863: Since you need the same number of entries as in `dataset`, and since the data is not really what makes the ops bogus (but the owner being diff. I think), I think you can just clone the `dataset` for the bogus ops.

src/sequence/mod.rs
```diff
@@ -719,22 +724,368 @@ mod tests {
...
+                        replica.apply_data_op(op)?;
+                    }
+                }
+            }
+
+            // TODO: Currently this wont converge. Data is missing.
+            // We need to be attempting to apply ops which are known to be not causally ready...
+            // We can get errors there by dropping from sync with one replica...
+            //
+            // Q: how to expand this across many replicas and simulate dropping messages? Is that wortwhile here at the data itself...?
```
bochaco: It seems we don't need to test that here as that would be part of upper layer's logic, like the Elders syncing up from time to time?
I think here at this layer we should just ensure that if all ops were applied in any order then the data converges, and as you did here perhaps also test that if any op is missing then they don't converge, etc.

src/sequence/mod.rs
```diff
@@ -719,22 +724,368 @@ mod tests {
...
             }
+
+            verify_data_convergence(replicas, dataset_length);
+        }
+
+
+
+        #[test]
+        fn proptest_dropped_data_can_be_reapplied_and_we_converge(
+            dataset in prop::collection::vec(generate_seq_entry(), 100..1001)
```
bochaco: What's the reason you want a minimum number of entries here instead of just using the `generate_dataset(1000)` ?

src/sequence/mod.rs
joshuef: I was thikning we needed different data in case the same data was somehow recognised as a no-op in the seq itself (through a bug there on the base seq detection)

src/sequence/mod.rs
joshuef: > It seems we don't need to test that here as that would be part of upper layer's logic, like the Elders syncing up from time to time?

Cools. I was thinking similar but wasn't 100% sure 👍

src/sequence/mod.rs
joshuef: in case we dropped everything, this makes that much less unlikely. But maybe isn't needed

src/sequence/seq_crdt.rs
```diff
@@ -278,6 +278,8 @@ where
                     Error::Unexpected("The data is an unexpected inconsistent state".to_string())
                 })?;
 
+                // FIXME: Check that we actually have perms to be adding a new policy here, based on prev one...
```
bochaco, line 281: We've done permissions check before calling any of this struct's functions: https://github.com/maidsafe/sn_data_types/blob/master/src/sequence/mod.rs#L264-L269, this struct/internal-layer doesn't know about permissions. Is there a scenario we are not covering yet?

bochaco (dismissed):
I think it's a good start what we have here, we should merge, the only thing is that comment I've just added (https://github.com/maidsafe/sn_data_types/pull/259#discussion_r520792385) where in theory there is a case not being covered.

S-Coyle (approved):

PR #269, opened 2020-11-16
feat(sequence): require signed ops

joshuef:


src/sequence/mod.rs
```diff
@@ -206,17 +206,17 @@ impl Data {
         })
     }
 
-    /// Appends new entry.
-    pub fn append(&mut self, entry: Entry) -> Result<DataWriteOp<Entry>> {
+    /// Generate unsigned crdt op, adding the new entry.
+    pub fn create_append_op(&mut self, entry: Entry) -> Result<DataWriteOp<Entry>> {
```
bochaco: We'll have to make a decision here, do we still want this function to apply the op to the replica or to only generate the operation? 
Currently `LSeq` applies it, thus `create_append_op` not only creates it but also applies it.

bochaco: Probably it's ok, as we don't want/need any signature verification for applying it to local replica??

joshuef: Hmmm, I was assuming we'd want it to not apply it. Seems cleaner? But also bring lseq in line with other crdt's in how they generate ops.

But you're right in that it's probably not _necessary_ for local data.

bochaco: Perhaps even for local replica we need the ops signed as for the event of a complete state replication, or e.g. when we will be sending old ops to other replica which is not causally ready?

joshuef: Yeh true. Where we save ops (for versioning, resending), will have to be at the `apply` stage (as no other point makes sense).

So yeh, will have to be signed. May be trivial to start saving ops there too now, and then as/when we want we can expose some kind of version api, with `version number` corresponding to the number of ops to apply (naiively at least... I suppose there's a Q of ordering a vec there too 🤔 )

src/sequence/mod.rs
```diff
@@ -409,20 +409,55 @@ mod tests {
...
     #[test]
     fn sequence_create_public() {
-        let actor = generate_public_key();
+        let actor1_keypair = Keypair::new_ed25519(&mut OsRng);
+        let actor = actor1_keypair.public_key();
         let sequence_name = XorName::random();
         let sequence_tag = 43_000;
-        let sequence = Sequence::new_public(actor, actor, sequence_name, sequence_tag);
+        let sequence =
+            Sequence::new_public(actor, "actor".to_string(), sequence_name, sequence_tag);
```
bochaco, line 460: Cosmetic here as it's for the test, but the actor (second arg.) could be `actor1_keypair.public_key().to_string()` instead of `"actor"`.
The first arg. shall be the PK this ops are being signed with, so perhaps change that var name to something else?

src/sequence/seq_crdt.rs
```diff
@@ -167,6 +174,18 @@ where
 
     /// Apply a remote data CRDT operation to this replica of the Sequence.
     pub fn apply_data_op(&mut self, op: CrdtDataOperation<A, Entry>) -> Result<()> {
+        // First check op is validly signed.
+        // Note: Perms for the op are checked at the upper Sequence layer.
+        match op.signature {
```
bochaco: Just in case you like this other way that I usually prefer:
```
let sig = op.signature.ok_or_else(|| Error::Unexpected("No signature on crdt op".to_string()))?;
...
op.source.verify(&sig, &bytes_to_verify)?;
...
```

bochaco (approved):

PR #271, opened 2020-11-17
Test failing - proptest converge with shuffled ops including bad ops which error and are not applied

S-Coyle:


S-Coyle:
superseded by PR #269

PR #289, opened 2020-12-07
feat(adult): add adult duties for chunk replication

Yoga07:


src/messaging/network.rs
```diff
@@ -54,17 +54,37 @@ pub enum NodeTransferCmd {
...
+        /// New Holder's name
         new_holder: XorName,
-        ///
+        /// Address of the blob to be duplicated
         address: BlobAddress,
-        ///
+        /// Current Holders
+        fetch_from_holders: BTreeSet<XorName>,
+    },
+    /// Get Chunk from current holders for duplication
```
joshuef, line 68: Is this _only_ for duplication purposes? ie. no blob requests would use this? Just wondering on naming here. It's maybe too broad atm?

joshuef (approved):

src/keys/signature.rs
```diff
@@ -48,6 +48,14 @@ impl Signature {
             _ => None,
         }
     }
+
+    /// Returns ed25519_dalek::Signature if Self is a Ed25519 variant.
+    pub fn into_ed(self) -> Option<ed25519_dalek::Signature> {
```
bochaco, line 53: Wouldn't/shouldn't this be implemented with the `From` trait?: https://doc.rust-lang.org/rust-by-example/conversion/from_into.html

Yoga07: We can do that for structs, but `Signature` is an enum here where it can be any one of the variants(BLS, Ed or BLSShare) therefore we have to `Option`-ize them to check and extract if it's a specific key as we see it to be. So we can't use the `From` trait here :)

PR #301, opened 2020-12-28
feat: use thiserror for errors

joshuef:


src/errors.rs
```diff
@@ -32,193 +33,130 @@ impl<'a, T> Debug for ErrorDebug<'a, T> {
 }
 
 /// Main error type for the crate.
-#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+#[derive(Error, Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub enum Error {
-    /// Access is denied for a given requester
+    /// Access denied for supplied PublicKey
+    #[error("Access Denied")]
     AccessDenied,
```
bochaco: Don't we want to provide the pk here so the consumer knows which pk was being used when the error was thrown? I know this is probably out of scope of this PR, but just thinking the direction we should take these errors to and start working on some consistency.

src/messaging/mod.rs
```diff
@@ -640,7 +640,7 @@ mod tests {
     }
 
     #[test]
-    fn try_from() -> Result<()> {
+    fn try_from() -> anyhow::Result<()> {
```
bochaco: Perhaps simply `use anyhow::{Result, anyhow}` for the whole `test` module?

joshuef: I did try that and things got wiiild w/ compiler errors. I'll double check that's not sorted now though

joshuef: that wildness is only in the seq file. the rest ill update

bochaco (approved):

PR #306, opened 2021-01-04
Feat remove messaging

joshuef:


src/map.rs
```diff
@@ -476,7 +476,7 @@ impl UnseqData {
         for (key, val) in insert {
             match new_data.entry(key) {
                 Entry::Occupied(entry) => {
-                    let _ = errors.insert(entry.key().clone(), EntryError::EntryExists(0));
+                    let _ = errors.insert(entry.key().clone(), Error::EntryExists(0));
```
Yoga07, line 479: Is it okay to have `let  _ =`s here? (there are a few below too).

src/lib.rs
```diff
@@ -56,17 +56,17 @@ pub use map::{
...
-    BlobWrite, Cmd, CmdError, DataAuthKind, DataCmd, DataQuery, Duty, ElderDuties, Event, MapRead,
-    MapWrite, Message, MessageId, MiscAuthKind, MoneyAuthKind, MsgEnvelope, MsgSender, NodeCmd,
-    NodeCmdError, NodeDataCmd, NodeDataError, NodeDataQuery, NodeDataQueryResponse, NodeDuties,
-    NodeEvent, NodeQuery, NodeQueryResponse, NodeRewardError, NodeRewardQuery,
-    NodeRewardQueryResponse, NodeSystemCmd, NodeTransferCmd, NodeTransferError, NodeTransferQuery,
-    NodeTransferQueryResponse, Query, QueryResponse, SequenceRead, SequenceWrite, TransferCmd,
-    TransferError, TransferQuery, TransientElderKey, TransientSectionKey, TryFromError,
-    MAX_LOGIN_PACKET_BYTES,
-};
+// pub use messaging::{
```
Yoga07: Can remove the commented lines.

src/errors.rs
```diff
@@ -145,17 +113,3 @@ pub enum Error {
 pub(crate) fn convert_bincode_error(err: bincode::Error) -> Error {
     Error::Serialisation(err.as_ref().to_string())
 }
-
-/// Entry error for `Error::InvalidEntryActions`.
-#[derive(Error, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
-pub enum EntryError {
```
Yoga07, line 151: Why are we removing this?

src/map.rs
joshuef: I wasn't looking at that in this PR. 

Seems to me like we're adding the error to a vec for later use though so should be fine

src/errors.rs
joshuef: We're putting them in the main Error struct now. So there's only one error type. It didn;t make sense to me to have two... Not sure why we did that?

Yoga07: It was specifically for the Sequence type's entry errors which were masked in `InvalidEntryActions` back then. But since we are moving towards a one main Error struct, I think removing this would make sense and add it to the main one.

Yoga07 (approved):

PR #313, opened 2021-01-26
Improve efficiency of sequence in_range

iancoleman:
This avoids iterating over every item in the lseq

### DIFF
--- a/src/sequence/seq_crdt.rs
+++ b/src/sequence/seq_crdt.rs
@@ -388,12 +388,13 @@ where
             return None;
         }
         let end_index = to_absolute_index(end, count)?;
+        let items_to_take = end_index - start_index;
 
         self.current_lseq().map(|lseq| {
             lseq.iter()
-                .enumerate()
-                .filter(|(i, _)| i >= &start_index && i < &end_index)
-                .map(|(_, entry)| entry.clone())
+                .skip(start_index)
+                .take(items_to_take)
+                .cloned()
                 .collect::<Entries>()
         })
     }


src/sequence/seq_crdt.rs
```diff
@@ -388,12 +388,13 @@ where
...
         }
         let end_index = to_absolute_index(end, count)?;
+        let items_to_take = end_index - start_index;
 
         self.current_lseq().map(|lseq| {
             lseq.iter()
-                .enumerate()
-                .filter(|(i, _)| i >= &start_index && i < &end_index)
-                .map(|(_, entry)| entry.clone())
+                .skip(start_index)
```
joshuef, line 395: Thanks for this @iancoleman !

One issue I see is that according to docs this will skip start index entirely. Whereas prior we were `>=`. 

I can see this failing for some client tests against node (to see same you have to have the whole local setup on the go just now). Looks like we're missing some tests here in sn_data_types.

`.skip(start_index -1)` gets it going for me w/ client. I'm happy to merge with that change in. But if you fancy writing a test to catch the current issue as well, that'd be amazing.

joshuef (changes requested):

src/sequence/seq_crdt.rs
joshuef: Hmmm, I think my startup section may have been broken unrelated to this. Rerunning the tests isn't giving same output again. Gonna put it through the ringer a few more times to double check, but may actually be fine.

joshuef: Aye,cannot repro. So this is good to go i think. Thanks @iancoleman

joshuef (approved):

PR #314, opened 2021-01-28
Rewards flow

oetyng:
Changes to enable the use of multisig transfer actor.


src/keys/keypair.rs
```diff
@@ -27,8 +27,6 @@ use threshold_crypto::{self, serde_impl::SerdeSecret};
 pub enum Keypair {
     /// Ed25519 keypair.
     Ed25519(ed25519_dalek::Keypair),
-    /// BLS keypair.
-    Bls(BlsKeypair),
```
Yoga07, line 31: Not sure why we are removing this?
(On a side note, we use BLS keypairs a lot at sn_client(esp for testing), removing this would induce a big refactor at client I think)

src/transfer.rs
```diff
@@ -25,6 +25,15 @@ pub type CreditId = [u8; 256 / 8];
 /// Msg, containing any data to the recipient.
 pub type Msg = String;
 
+///
```
joshuef: missing docs here

joshuef (changes requested):

Yoga07 (dismissed):

joshuef (approved):

PR #317, opened 2021-02-01
Feat signing+ownershiip

joshuef:


src/keys/keypair.rs
```diff
@@ -62,21 +138,21 @@ impl Keypair {
     /// Constructs a random Ed25519 public keypair.
```
dirvine, line 138: Be good to remove the word public in the comment.

dirvine (approved):
LGTM, 1 small comment

Yoga07 (approved):

PR #324, opened 2021-02-02
Omit address when serializing private data

davidrusu:
Depends on https://github.com/maidsafe/xor_name/pull/59

`PrivateKey::new()` now no longer returns result since we don't depend on bincode to derive the address.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ crdts = "4.3.0"
 sha3 = "~0.8.2"
 threshold_crypto = "~0.4.0"
 tiny-keccak = "~1.5.0"
-xor_name = "1.1.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -19,7 +19,7 @@ crdts = "4.3.0"
 sha3 = "~0.8.2"
 threshold_crypto = "~0.4.0"
 tiny-keccak = "~1.5.0"
-xor_name = "1.1.0"
+xor_name = { git = "https://github.com/maidsafe/xor_name", rev="72f1c1a02101db4612cc8e8d9aa64e490034b938" }
```
davidrusu: Will update this once we've merged https://github.com/maidsafe/xor_name/pull/59

madadam (dismissed):

Yoga07 (dismissed):

Yoga07 (approved):

PR #335, opened 2021-02-17
Update Sequence to use List CRDT

davidrusu:
Based on the work in https://github.com/rust-crdt/rust-crdt/pull/104

@iancoleman You may want to take a look at this, it replaces the internal identifier representation in LSeq with an arbitrary precision rational number. The code no longer panics after too many inserts.


src/sequence/mod.rs
```diff
@@ -2331,177 +2218,6 @@ mod tests {
 
             // now we converge
             verify_data_convergence(vec![replica1, replica2], dataset_length)?;
-
-        }
-
-
-        #[test]
-        fn proptest_dropped_data_can_be_reapplied_and_we_converge(
```
davidrusu, line 1083: These tests here are not valid since the List (and LSeq) CRDT's expect source ordering. We can't drop or shuffle ops arbitrarily and expect these CRDT's to converge unfortunately.

dan-da: > the List (and LSeq) CRDT's expect source ordering

Has this (somehow) always been the case, or the source ordering requirement is a result of the changes in this PR?

src/sequence/mod.rs
```diff
@@ -533,9 +517,9 @@ mod tests {
         )?;
         replica1.apply_data_op(op2.clone())?;
 
-        // we apply the operations in different order, to verify that doesn't affect the result
-        replica2.apply_data_op(op2)?;
+        // Op's must be applied in the same order a source created them (Source Ordering)
```
bochaco: Isn't this problematic? what happens if the op when broadcasted reaches a replica in different order? is this now a limitation in crdts::List?, I wouldn't have expected it to be.

src/sequence/mod.rs
bochaco: I'm concerned about this too, isn't this limitation a big problem for us? I don't see why this limitation now exists as from my understanding this didn't in previous LSeq impl, the only problem was the mutability of Policies, but since we are removing that then I wouldn't have expected ordering to be a problem?

davidrusu: It did previously as well, it just wasn't enforced.

Deletes break commutativity,  that is `Insert(3, "A"), Delete(3)` is not the same as `Delete(3), Insert(3, "A")`. You've gotten away with it in the past because LSeq was not doing idempotency checks + you don't yet support deletes.

Idempotency checks are needed again when you have deletes, consider the situation: `Insert(3, "A"), Delete(3)` followed by the originial `Insert(3, "A")` being replayed. You'll find that "A" re-appears  in the list.

With the idempotency checks in place, the replayed insert is a no-op.

> isn't this limitation a big problem for us?

It shouldn't be, we need to check that sequence numbers are strictly incrementing before applying an operation. Otherwise you error back and ask the sender to retry and include the missing operations.

bochaco: I see, we don't allow delete ops in public data, so perhaps we can get away with it and not require the order. For private data we still need to define if we will allow it or not, ideally we don't want different behaviors/logic/checks between private and public though....

dan-da: > With the idempotency checks in place, the replayed insert is a no-op.

So it sounds like you are saying that LSeq can be made idempotent and therefore support Deletes, but that work has not been done yet.  Plus it requires source ordering.  Is that right?

So to summarize then:

1. We can have an LSeq without source ordering, but it cannot support deletes.
2. We can have an LSeq with Deletes, but it requires source ordering for correctness (to be idempotent).
3. Source ordering requirement is not a big deal for us because we can always ask source to send missing ops.

Have I stated that correctly?

If so, do we need to add some functionality for (3) into BRB?   Something like combine MsgDotNotTheNextDot error with an anti_entropy request?

iancoleman:
> The code no longer panics after too many inserts.

Nice. The List code looks very tidy and much simplified over LSeq. But as for the details, my knowledge is not deep enough there and the previous comments are more useful than anything I'd add.

I ran a test including serialization which closely simulates a client/node interaction, and it passes no worries.

```rust
#[test]
fn test_append_many() {
    let mut client = List::new();

    // client serializes the list for storage on the node
    let client_serialized = rmp_serde::to_vec(&client).unwrap();

    // node receives the serialized list from the client
    let mut node_serialized = client_serialized.clone();

    // client uploads the list to the node
    let mut node: List<String, SiteId> = rmp_serde::from_read_ref(&node_serialized).unwrap();

    // node serializes for storage
    node_serialized = rmp_serde::to_vec(&node).unwrap();

    let appends = 10000;
    for i in 0..appends {
        // client gets the list over the wire from serialized node storage
        client = rmp_serde::from_read_ref(&node_serialized).unwrap();
        // client appends a new value
        let val = format!("value {}", i);
        let op = client.append(val, 0);
        client.apply(op.clone()); // optional?
        // serialize op to send to node
        let op_serialized = rmp_serde::to_vec(&op).unwrap();
        // node deserializies op
        let op_deserialized: Op<String, SiteId> = rmp_serde::from_read_ref(&op_serialized).unwrap();
        // node applies op to the node list
        node.apply(op_deserialized);
        // node serializes their new list for storage
        node_serialized = rmp_serde::to_vec(&node).unwrap();
    }

    assert_eq!(client.len(), appends);
    assert_eq!(node.len(), appends);
}
```

dirvine:
Excellent @iancoleman 

@davidrusu 
In terms of Delete requiring order, I agree, but for this reason and to ensure we remain a semi-lattice CRDT type I would not allow deletes. 

I really think we need to consider data types as CRDT (semi-lattice) or not. 

However even CRDT types were we deterministically resolve forks, then we need to be certain forks are not harmful (i.e. doublespend). Forks created by mistake or malice for CRDT is no worry, as long as the fork has no side effects. CRDT is OK there. 

Where we cannot allow forks (i.e. crypto currency) then we must enforce an order **co-ordinated order** to prevent the creation of a fork. So we need a clear ruleset here of data and the capabilities (or not) of such data.

src/sequence/mod.rs
davidrusu: @dan-da that sounds about right, 

> If so, do we need to add some functionality for (3) into BRB? Something like combine MsgDotNotTheNextDot error with an anti_entropy request?

Not necessary I would say, the eventual anti-entropy rounds should take care of this for us.

bochaco (dismissed):

bochaco (approved):

PR #346, opened 2021-02-23
Switch to glist for seq

bochaco:


Cargo.toml
```diff
@@ -15,7 +15,8 @@ bincode = "1.2.1"
 multibase = "~0.8.0"
 hex_fmt = "~0.3.0"
 rand = "~0.7.3"
-crdts = "4.3.0"
+crdts = { git = "https://github.com/rust-crdt/rust-crdt.git", branch = "master" }
```
davidrusu, line 18: 6.2.0 has been released with GList

src/sequence/mod.rs
```diff
@@ -179,297 +152,140 @@ impl Data {
...
-        Ok(match &self.data {
+        let items = match &self.data {
             SeqData::Public(data) => data.in_range(start, end),
             SeqData::Private(data) => data.in_range(start, end),
-        })
+        };
+
+        let entries = match items {
+            None => None,
+            Some(entries) => Some(entries.iter().map(|(_, entry)| entry.to_vec()).collect()),
```
davidrusu, line 164: I think we should be exposing the Markers here to the application to avoid race conditions when appending data

src/sequence/mod.rs
```diff
@@ -179,297 +152,140 @@ impl Data {
...
     /// Returns a value at 'index', if present.
     pub fn get(&self, index: Index, requester: Option<PublicKey>) -> Result<Option<&Vec<u8>>> {
-        self.check_permission(Action::Read, requester, None)?;
+        self.check_permission(Action::Read, requester)?;
 
         Ok(match &self.data {
-            SeqData::Public(data) => data.get(index),
-            SeqData::Private(data) => data.get(index),
+            SeqData::Public(data) => data.get(index).map(|(_, entry)| entry),
+            SeqData::Private(data) => data.get(index).map(|(_, entry)| entry),
```
davidrusu, line 176: Again, clients need to use this marker if they want to append after this entry

src/sequence/mod.rs
```diff
@@ -179,297 +152,140 @@ impl Data {
...
+    /// Returns the last entry, if it's not empty.
     pub fn last_entry(&self, requester: Option<PublicKey>) -> Result<Option<&Entry>> {
-        self.check_permission(Action::Read, requester, None)?;
+        self.check_permission(Action::Read, requester)?;
 
         Ok(match &self.data {
-            SeqData::Public(data) => data.last_entry(),
-            SeqData::Private(data) => data.last_entry(),
+            SeqData::Public(data) => data.last_entry().map(|(_, entry)| entry),
+            SeqData::Private(data) => data.last_entry().map(|(_, entry)| entry),
```
davidrusu, line 186: Another dropped marker, I'll stop commenting on this now :)

PR #350, opened 2021-03-02
feat(Seq): making the Policy of a Sequence data type immutable

bochaco:
BREAKING CHANGE: Policy mutation APIs are removed and Sequence constructors need the Policy to be provided.


davidrusu (approved):
:ship: :fire:

PR #352, opened 2021-03-03
Adding Register data type

bochaco:


Cargo.toml
```diff
@@ -15,14 +15,15 @@ bincode = "1.2.1"
 multibase = "~0.8.0"
 hex_fmt = "~0.3.0"
 rand = "~0.7.3"
-crdts = "4.3.0"
+crdts = { git = "https://github.com/rust-crdt/rust-crdt.git", branch = "master" }
+#crdts = "6.3.0"
```
davidrusu: 6.3.1 should have what you need

src/register/reg_crdt.rs
```diff
@@ -0,0 +1,139 @@
...
+        // Let's first check the op is validly signed.
+        // Note: Perms for the op are checked at the upper Register layer.
+        let sig = op.signature.ok_or(Error::CrdtMissingOpSignature)?;
+        let bytes_to_verify = utils::serialise(&op.crdt_op).map_err(|err| {
+            Error::Serialisation(format!(
+                "Could not serialise CRDT operation to verify signature: {}",
+                err
+            ))
+        })?;
+        op.source.verify(&sig, &bytes_to_verify)?;
```
davidrusu, line 116: Should probably also validate that `op.address == self.address`

src/register/reg_crdt.rs
```diff
@@ -0,0 +1,139 @@
...
+        Ok(())
+    }
+
+    /// Gets the entry corresponding to the provided `hash` if it exists.
+    pub fn get(&self, hash: EntryHash) -> Option<&Entry> {
+        self.data.node(hash).map(|node| &node.value)
+    }
+
+    /// Gets the last entry, or entries if there are branches.
+    pub fn last(&self) -> BTreeSet<(EntryHash, Entry)> {
```
davidrusu: The Register CRDT family has a few common interfaces, we should follow one of them instead of using our own concepts here (i.e. `last`):

The one rust-crdt uses:
```
Register:
  write(values, ctx)
  read()
```

Other common ones:
```
Register:
  update(val, ctx)
  query()
```
```
Register:
  assign(val, ctx)
  value()
```

I'm not religious about which one to pick, but I'd prefer we follow one of the common conventions.

src/register/reg_crdt.rs
```diff
@@ -0,0 +1,139 @@
...
+        // FIXME: if there are multiple branches, resolve which is the
```
davidrusu: can probably remove this comment now

davidrusu (dismissed):
Looks good :+1:

davidrusu (dismissed):
Looks good

S-Coyle (dismissed):

PR #353, opened 2021-03-11
feat: adds hex-parse for user-facing PK/SK variants + hex fmt for PK

### LINKED ISSUE #347: Exporting hex_string<->bytes conversion utilities
There are several consumers of `sn_data_types` that currently use some version of hex_string<->`Vec<u8>` conversion. This includes `sn_node`, `sn_api`, and `sn_routing` to name a few. All of these either re-implement functions like `parse_hex` or `bytes_to_hex` or pull in the `hex` or `hex_fmt` crates. 

Usually, this is done in dealing with public/private keys exposed by this crate. Example use below:

```
Let bytes = hex_to_vec(cli_args.key);
Let key = PublicKey::Ed25519(
    ed2559_dalek::from_bytes(bytes)?
);
```

Some reimplementation/redundant sites I found:
 * [sn_node](https://github.com/maidsafe/sn_node/blob/798fd7a6bca92210da295dd40bb8bcf107076e1d/src/node/state_db.rs#L59)
 * [sn_api](https://github.com/maidsafe/sn_api/blob/9c2efaaf564ced1130b8be4faab34e8b090c5885/sn_api/src/api/common.rs#L64)
 * [sn_message](https://github.com/maidsafe/sn_messaging/blob/5a9d580ad36a699403a1aafa6b26a89e97614de2/src/client/sender.rs#L52)(Actually, this isn't reimplementation as it is importing `HexFmt` when `PublicKey` could also implement `fmt::LowerHex` or just be converted to hex before display)

Consumers of `sn_data_tyeps` could benefit from reduced code redundancy and easier dependency management if `sn_data_types` re-exported the `hex` crate to expose functions like `encode` and `decode` that are used or re-implemented in the consumers of `sn_data_types`.

Alternatively, a convenience function could be added to `utils.rs` file to expose `vec_to_hex` and `hex_to_vec` functions so that consumers don't need to worry about introducing a new `HexFormatError` type exposed by the `hex` crate if we were to re-export `hex` directly.

I can throw together a PR in a bit to support this concept with a concrete implementation. But, for now, I'll put this git issue here to present the idea.

Scorch-Dev:
Based on [this forum discussion](https://forum.safedev.org/t/sn-data-types-publickey-wrapper-usage/2972/10), I agree that it makes sense to start implementing hex-parse functionality for key types ahead of time, and then slowly start diffusing the changes to other crates as they become more stable. This would still allow us to eventually cut down on dependencies while still maintaining stability as needed.

Closes #347 

Adds:

-- PublicKey::ed_from_hex
-- PublicKey::bls_from_hex
--SecretKey::ed_from_hex

Refactors:

--Moves `hex` crate from `dev-dependencies` to `dependencies`
--Removes (unmaintained?) `hex_fmt` crate from `dependencies`


joshuef:
Ah nice one @Scorch-Dev ! 

Sorry this seems to have fallen under the radar the last week or two. We're getting some larger node changes (some simplification) in, and then I;ll loop back around to give this a proper 👀 .

Hope that's all right?

src/keys/secret_key.rs
```diff
@@ -26,6 +27,24 @@ pub enum SecretKey {
...
+impl SecretKey {
+    /// Construct a secret key from a hex string
+    ///
+    /// Similar to public key, it is often useful in user
+    /// facing apps to be able to set your own secret
+    /// key without depending on both the ed25519_dalek
+    /// and hex crates just to reimplement this function
+    pub fn ed25519_from_hex(hex: &str) -> Result<Self> {
+        let bytes = hex::decode(hex).map_err(|_| {
+            Error::FailedToParse("Couldn't parse bytes from secret key hex".to_string())
```
bochaco: I think we can include the `hex::decode` error in the `FailedToParse` message, and do this consistently in the other cases, including for the ones for public key. I like this one which has more context info since it's saying what has failed exactly when trying to decode it as hex, along with the hex decoding error.

src/keys/public_key.rs
```diff
@@ -191,6 +226,24 @@ impl Display for PublicKey {
     }
 }
 
+impl LowerHex for PublicKey {
+    /// Useful for displaying public key in user-facing apps
+    /// E.g. in cli and in human-readable messaging like for sn_authd
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        let bytes = self.clone().to_bytes();
```
bochaco: I think we can get rid of cloning and just call `to_bytes()`

bochaco (changes requested):
This looks good to me @Scorch-Dev , just a couple of minor things I'd improve there

Scorch-Dev:
Yea no worries, I get the impression it’s busy right now. As always, thanks for the feedback. It all sounds good to me so I’ll patch those in later today 👌

bochaco (approved):
Thanks @Scorch-Dev , this looks good to me.

PR #363, opened 2021-04-27
feat(api): add DataAddress enum for address of different data types

lionel-faber:


src/lib.rs
```diff
@@ -90,6 +90,19 @@ pub enum Data {
     Register(Register),
 }
 
+/// Object storing an address of data on the network
+#[derive(Clone, Eq, PartialEq, Hash, Serialize, Deserialize, Debug, PartialOrd, Ord)]
```
Yoga07, line 94: Do we want to derive `PartialOrd` and `Ord` traits?

lionel-faber: Yes we do. It's required [here](https://docs.rs/sn_messaging/19.0.0/sn_messaging/client/enum.Error.html#impl-Ord)

Yoga07 (approved):

